Issue #7, opened 2020-06-09
Move to Typescirpt


mihirsoni:
For better maintainability we started running most of the plugins to typescript. Kibana itself has moved 100 % typescript. You can refer AD / ISM

PR #14, opened 2020-06-17
Migrate public/main to Typescript

davidcui-amzn:
*Issue #, if available:*
#7 
*Description of changes:*
Migrate all of `public/main` into Typescript. This includes the homepage view of the Reporting plugin page. 


public/components/main/reports_table.tsx
```diff
@@ -34,7 +35,7 @@ export const reports_list_columns = [
   {
     field: 'reportSource',
     name: 'Source',
-    render: username => (
+    render: (username: any) => (
```
mihirsoni, line 38: Shouldn't this be string ?

public/components/main/reports_table.tsx
```diff
@@ -106,21 +103,27 @@ const renderToolsLeft = () => {
...
+};
+
 export const reports_list_selection_value = {
-  selectable: user => user.online,
-  selectableMessage: selectable =>
+  selectable: (user: { online: any; }) => user.online,
+  selectableMessage: (selectable: any) =>
       !selectable ? 'User is currently offline' : undefined,
-  onSelectionChange: selection => setSelection(selection),
+  onSelectionChange: (selection: any) => onSelectionChange,
```
seraphjiang, line 128: try use specific type
```
seleactable:boolean
```

public/components/main/reports_table.tsx
davidcui-amzn: Yes, will change

public/components/main/reports_table.tsx
davidcui-amzn: Will change

public/components/main/scheduled_reports_table.tsx
```diff
@@ -33,7 +34,7 @@ export const scheduled_report_columns = [
   {
     field: 'reportSource',
     name: 'Source',
-    render: username => (
+    render: (username: any) => (
```
zhongnansu, line 37: should this be `string`?

davidcui-amzn: good catch, will change

PR #17, opened 2020-06-17
Migrate create_report/ to Typescript

davidcui-amzn:
*Issue #, if available:*
#7 
*Description of changes:*
Migrate all files in `create_report/` to Typescript. All UI will be migrated to Typescript after this PR and #14 are merged.


public/components/create_report/create_report.tsx
```diff
@@ -38,14 +38,34 @@ import {
...
+  reportSettingsRadioIdSelected: string,
+  reportSettingsSetRadioIdSelected: string,
+  reportSettingsDashboard: string,
+  deliveryEmailSubject: string,
+  deliveryEmailBody: string,
+  scheduleRadioFutureDateSelected: boolean,
+  scheduleRadioRecurringSelected: boolean,
+  scheduleUTCOffset: number,
+  scheduleRecurringFrequency: string,
+  scheduleRecurringDailyTime: string,
```
mihirsoni, line 62: Shouldn't this be number ? Is it ISO string or timestamp ?

public/components/create_report/create_report.tsx
```diff
@@ -38,14 +38,34 @@ import {
...
+  scheduleRadioRecurringSelected: boolean,
+  scheduleUTCOffset: number,
+  scheduleRecurringFrequency: string,
+  scheduleRecurringDailyTime: string, 
+  scheduleRecurringUTCOffset: number,
+  scheduleRecurringWeeklyDayOfWeek: string,
+  scheduleRecurringStartDate: Moment
+}
+
+export class CreateReport extends React.Component<RouterHomeProps, CreateReportState> {
```
mihirsoni, line 68: np: Use functional components and hooks make things much more easy to write no more overhead on classes. 

* https://reactjs.org/docs/hooks-intro.html 
You can take look at AD plugin as well.

public/components/create_report/create_report.tsx
davidcui-amzn: Looks like this variable isn't even used, removing it altogether

public/components/create_report/create_report.tsx
```diff
@@ -0,0 +1,310 @@
...
+import React, { useState } from 'react';
+import {
+    EuiButtonEmpty,
+    EuiDatePicker,
+    EuiSelect,
+    EuiFlexGroup,
+    EuiFlexItem,
+    EuiFormRow,
+    EuiButton,
+    EuiPage,
```
zhongnansu, line 25: Noticed that you added `@ts-ignore` to the same component `EuiPage` in this PR, https://github.com/opendistro-for-elasticsearch/kibana-reports/pull/14. What's the reason for different style

davidcui-amzn: Got an error in my IDE that `EuiPage` didn't exist even though everything was functioning fine, so I added it to silence the warning

public/components/create_report/create_report.tsx
```diff
@@ -0,0 +1,310 @@
...
+  const [deliveryEmailBody, setDeliveryEmailBody] = useState('');
+  const [scheduleRadioFutureDateSelected, setScheduleRadioFutureDateSelected] = useState(false);
+  const [scheduleRadioRecurringSelected, setScheduleRadioRecurringSelected] = useState(false);
+  const [scheduleUTCOffset, setScheduleUTCOffset] = useState(0);
+  const [scheduleFutureDate, setScheduleFutureDate] = useState(moment());
+  const [scheduleRecurringFrequency, setScheduleRecurringFrequency] = useState('Daily');
+  const [scheduleRecurringUTCOffset, setScheduleRecurringUTCOffset] = useState(0);
+  const [scheduleRecurringWeeklyDayOfWeek, setScheduleRecurringWeeklyDayOfWeek] = useState('Monday');
+  const [scheduleRecurringStartDate, setScheduleRecurringStartDate] = useState(moment());
+  const [scheduleRadioIdSelected, setScheduleRadioIdSelected] = useState(`${idPrefix}7`);
```
mihirsoni, line 51: Use can also use `useState` with object. Such as 

```
const reportState = useState<CreateReportState>({
....otherstates.
})
```

public/components/create_report/create_report.tsx
```diff
@@ -0,0 +1,310 @@
...
+  const ScheduleRecurringWeeklyInput = () => {
+    const recurringDayOptions = [
+      { value: 'Monday', text: 'Monday' },
+      { value: 'Tuesday', text: 'Tuesday' },
+      { value: 'Wednesday', text: 'Wednesday' },
+      { value: 'Thursday', text: 'Thursday' },
+      { value: 'Friday', text: 'Friday' },
+      { value: 'Saturday', text: 'Saturday' },
+      { value: 'Sunday', text: 'Sunday' },
+    ];
```
mihirsoni, line 185: np: You could move this outside of the component as it is constant.

public/components/create_report/create_report.tsx
```diff
@@ -0,0 +1,310 @@
...
+
+    const onChangeDayOfWeek = (e) => {
+      setScheduleRecurringWeeklyDayOfWeek(e.target.value);
+    }
```
mihirsoni, line 189: `e: React.MouseEvent<>` missing type.

public/components/create_report/create_report.tsx
```diff
@@ -0,0 +1,310 @@
...
+      ? <ScheduleRecurringDailyInput/>
+      : null;
+
+    const weekly = (scheduleRecurringFrequency == 'Weekly')
+      ? <ScheduleRecurringWeeklyInput/>
+      : null;
+
+    const monthly = (scheduleRecurringFrequency == 'Monthly')
+      ? <ScheduleRecurringMonthlyInput/>
+      : null;
```
mihirsoni, line 245: Also always have best practice to move use constants. 
Simplify with switch.
```ts
const frequency = null;

switch(scheduleRecurringFrequency: string)  {
 case 'Daily': 
        frequencyContent = <ScheduleRecurringDailyInput/>
       break;
  case 'Weekly':
        frequencyContent = <ScheduleRecurringWeeklyInput/>
        break;
    default:
    break;
}

File: public/components/create_report/create_report.tsx (hunk 6 above)
Comment (davidcui-amzn) [reply public/components/create_report/create_report.tsx:189]: Shouldnt this be `string` since the options are the days of the week which are strings?

File: public/components/create_report/create_report.tsx (hunk 4 above)
Comment (davidcui-amzn) [reply public/components/create_report/create_report.tsx:61]: Would I be defining the other states in `CreateReportState`? Otherwise I don't think I fully understand the added benefit

Comment (mihirsoni) [reply public/components/create_report/create_report.tsx:61]: Correct you can create a State which contains all the required info and move this inside une useState. to simplify

File: public/components/create_report/create_report.tsx [hunk 8]
```diff
@@ -0,0 +1,309 @@
...
+  const timezone_options = [
+    { value: -4, text: 'EDT -04:00' },
+    { value: -5, text: 'CDT -05:00' },
+    { value: -6, text: 'MDT -06:00' },
+    { value: -7, text: 'MST/PDT -07:00' },
+    { value: -8, text: 'AKDT -08:00' },
+    { value: -10, text: 'HST -10:00' }
+  ];
+
+  const ScheduleFutureDatePicker = () => {
```
mihirsoni, line 125: You could move all individual components out to simplify and better readability.

public/components/create_report/create_report.tsx
mihirsoni: No the `e` should be of type event.

Issue #25, opened 2020-06-25
Change to use streaming for report download API


zhongnansu:
ref: #11
The existing solution for file download is by first storing a file on server, and then send to client. 
Since the storage of file is unnecessary, we decided to switch to use streaming to send file(pdf/png)

PR #30, opened 2020-06-29
Stream reports to client

zhongnansu:
*Issue #, if available:*
#21 #25 
*Description of changes:*
- Major
  - Stream reports to client to git rid of storing actual file on server
- Minor
  - Added config file for `eslint`, `prettier` and `kibana-plugin-helpers.json`
  - init commit on adding API to list all reports and single report(by its id)


.lintstagedrc
```diff
@@ -0,0 +1,3 @@
+{
+  "*.{ts,tsx,js,jsx,json,css,md}": ["prettier --write", "git add"]
+}
```
mihirsoni, line 3: np: new line

package.json
```diff
@@ -21,17 +22,17 @@
...
     "@testing-library/jest-dom": "^5.10.1",
     "@testing-library/react": "^10.2.1",
     "@types/jest": "^26.0.0",
     "@types/node": "^14.0.13",
+    "@types/reach__router": "^1.3.5",
     "@types/react": "^16.9.36",
     "@types/react-dom": "^16.9.8",
     "@types/react-test-renderer": "^16.9.2",
-    "images-to-pdf": "^1.0.3",
+    "gensync": "^1.0.0-beta.1",
```
mihirsoni, line 35: Ensure we don't get any breaking changes and use this on production.

server/services/generateReportService.ts
```diff
@@ -14,99 +14,208 @@
...
         itemName,
+        source,
         reportFormat,
         windowWidth = 1200,
         windowLength = 800,
       } = req.payload as {
         url: string;
         itemName: string;
+        source: string;
         reportFormat: string;
```
mihirsoni, line 55: Create whitelisting for `reportFormat` to ensure no other formats are supported or being accepted from server end. Client can manipulate anything.

server/services/generateReportService.ts
```diff
@@ -14,99 +14,208 @@
...
-        windowWidth: number;
-        windowLength: number;
+        windowWidth?: number;
+        windowLength?: number;
```
mihirsoni, line 57: Do we need any specific validation our widht and height ? Such as some ration to ensure we don't mess the report ?

server/services/generateReportService.ts
```diff
@@ -14,99 +14,208 @@
...
+        return (
+          h
+            .response(stream)
+            .type('application/pdf')
+            .header('Content-type', 'application/pdf')
+            //@ts-ignore
+            .header('Content-length', stream.length)
+            .header(
+              'Content-Disposition',
+              `attachment; filename=${fileName}.${reportFormat}`
```
mihirsoni, line 111: While streaming will there be any other side effect such as time out  ? size issue ?

server/services/generateReportService.ts
zhongnansu: I assume you mean input validation. I have done that with `Joi` here https://github.com/opendistro-for-elasticsearch/kibana-reports/blob/a9c122fcf20f57ce48c4675cca6089e1eae4ac69/server/routes/generateReport.ts#L37

server/services/generateReportService.ts
zhongnansu: Good point. I plan to use a maximum and minimum value to restrict the width and height. any idea how to figure out that value?

.lintstagedrc
zhongnansu: will do

package.json
zhongnansu: this dependency was supposed to be removed. will remove

server/services/generateReportService.ts
davidcui-amzn: Puppeteer has some pre-set page dimensions that can be used as reference- if not that, then maybe using default width and height for web pages?

server/utils/constants.ts
```diff
@@ -16,6 +16,8 @@
 export const BASE_API_PATH = '/api/reporting';
 export const NODE_API = Object.freeze({
   GENERATE_REPORT: `${BASE_API_PATH}/generateReport`,
+  GET_REPORTS: `${BASE_API_PATH}/reports`,
+  GET_REPORT: `${BASE_API_PATH}/reports/{reportId}`,
```
davidcui-amzn, line 20: How do you feel about changing this to something like `GET_SINGLE_REPORT`? `GET_REPORT` vs `GET_REPORTS` seems too similar to me

zhongnansu: How does `GET_REPORT_BY_ID` sound to you?

server/services/generateReportService.ts
zhongnansu: 1. Puppeteer default viewport size is 800 * 600 px. I think what we are discussing here is the max/min, not the default value. In my design, I set the default to be 1200 * 800, if no input is given.
2. Yeah I am thinking what's the max/min resolution for responsive web pages. Some knowledge or design convention from front-end is needed. E.g., if a user gives an input 100 * 100 px, or 10000* 10000 px, which we can't allow to happen. Clearly puppeteer doesn't have restriction on that. The image here shows the png generated using a 100 * 100 px setting.
image

server/services/generateReportService.ts
zhongnansu: 1. For timeout, do you mean I should set a timeout on server side, such as a socket timeout?
2. For size, it refers to the size of the report pdf/png file(screenshot). I think no matter how large that screenshot will be, we should generate it and deliver to user. Because it's provided by user on the kibana dashboard or kibana visualization tab.

server/utils/constants.ts
davidcui-amzn: 👍

server/services/generateReportService.ts
mihirsoni: Got it. Thanks

Issue #6, opened 2020-06-09
Move plugin to new kibana platform.


mihirsoni:
We should move this kibana plugin to latest platform and no support for legacy.

zhongnansu:
server    log   [05:20:51.457] [warning][legacy-service] Some installed third party plugin(s) [example_plugin] are using the legacy plugin format and will no longer work in a future Kibana release. Please refer to https://ela.st/kibana-breaking-changes-8-0 for a list of breaking changes and https://ela.st/kibana-platform-migration for documentation on how to migrate legacy plugins.

PR #41, opened 2020-07-11
Migrate Client-Side to New Platform

davidcui-amzn:
*Issue #, if available:*
#6 
*Description of changes:*
* update Kibana version to `8.0.0` 
* new plugin definition: deleted `router_home.tsx` and replaced functionality in `application.tsx`
* update snapshot tests
* enable UI in `kibana.json` 


public/components/app.tsx
```diff
@@ -0,0 +1,107 @@
...
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import React from 'react';
+import { FormattedMessage, I18nProvider } from '@kbn/i18n/react';
+import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
```
seraphjiang, line 18: there is some issue to use BrowserRouter
in case observers some  route logic doesn't meet expectation,  while adding more page, and more complex navigation between page. 

we may switch to HashRouter, not require to change, just FYR

public/plugin.ts
```diff
@@ -0,0 +1,73 @@
...
+    return {
+      getGreeting() {
+        return i18n.translate('opendistroKibanaReports.greetingText', {
+          defaultMessage: 'Hello from {name}!',
+          values: {
+            name: PLUGIN_NAME,
+          },
+        });
+      },
+    };
```
seraphjiang, line 65: remove this, just ``` return {};```

public/types.ts
```diff
@@ -0,0 +1,26 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import { NavigationPublicPluginStart } from '../../../src/plugins/navigation/public';
+
+export interface OpendistroKibanaReportsPluginSetup {
+  getGreeting: () => string;
```
seraphjiang, line 19: remove getGreeting

public/components/app.tsx
davidcui-amzn: Will switch- was using Hash router in old platform and don't want to incur any avoidable tech debt

public/plugin.ts
davidcui-amzn: Will do

public/types.ts
davidcui-amzn: will remove

Issue #46, opened 2020-07-20
CRUD APIs for report definition


zhongnansu:
APIs for report definition should enable caller to:
- create a report definition
- update a report definition
- get all report definitions
- get single report definition by id
- delete a report definition

Issue #45, opened 2020-07-20
CRUD APIs for reports


zhongnansu:
report APIs should be able to:
- stream report to client(already have)
- get all reports details
- get single report detail by id
- delete a report
- no update in our scenario

Issue #20, opened 2020-06-23
Data modeling in Elasticsearch


zhongnansu:
- Metadata
  - Save metadata
  - Update/Retrieve matadata

PR #47, opened 2020-07-20
CRUD APIs for report and report definition

zhongnansu:
*Issue #, if available:*
#20 #45 #46 
*Description of changes:*
- Added schema for report/report definition
- report API:
  - GET `api/reporting/reports?size=5&sortField=time_created&sortDirection=desc`
  - GET `api/reporting/reports/{reportId}`
  - DELETE `api/reporting/reports/{reportId}`

- report definition API:
  - POST `api/reporting/reportDefinition`
  - GET `api/reporting/reportDefinitions?sortField=time_created&sortDirection=desc&size=5`
  - GET `api/reporting/reportDefinitions/{reportDefinitionId}`
  - PUT `api/reporting/reportDefinitions/{reportDefinitionId}`
  - DELETE `api/reporting/reportDefinitions/{reportDefinitionId}`


server/model/index.ts
```diff
@@ -0,0 +1,117 @@
...
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+import { schema } from '@kbn/config-schema';
+
+// TODO: needs update when we integrate csv feature
+const data_report_params = schema.object({
```
davidcui-amzn, line 19: np: constants should be in all caps

davidcui-amzn, line 19: how come these consts are snake case and the others in the file are camel case?

server/model/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  has_attachment: schema.boolean({ defaultValue: true }),
+  recipients: schema.arrayOf(schema.string(), { minSize: 1 }),
+});
+
+export const reportSchema = schema.object({
+  report_name: schema.string(),
+  report_source: schema.oneOf([
+    schema.literal('Dashboard'),
+    schema.literal('Visualization'),
+    schema.literal('Saved Search'),
```
davidcui-amzn, line 75: From the UX this entry is listed as `Saved search` vs. `Saved Search`, the search is not capitalized

server/model/index.ts
```diff
@@ -0,0 +1,117 @@
...
+  report_params: schema.conditional(
+    schema.siblingRef('report_source'),
+    'Saved Search',
+    data_report_params,
+    visual_report_params
+  ),
+
+  delivery: schema.maybe(
+    schema.object({
+      channel: schema.oneOf([
```
davidcui-amzn, line 92: Since we only have Kibana and Email as options right now, I think it should be ok to leave out Slack and Chime

zhongnansu: That's true, but it doesn't affect the schema definition here

server/model/index.ts
zhongnansu: will change

server/model/index.ts
zhongnansu: thx for pointing that out. Changed the `consts` to use camel case

server/routes/report.ts
```diff
@@ -214,9 +208,7 @@ export default function (router: IRouter) {
           }
         );
         return response.ok({
-          body: {
-            data: esResp,
-          },
+          body: esResp._source,
```
davidcui-amzn, line 211: question: does this change the response returned from the API at all? Or is it refactoring for more elegant API work in back-end?

server/routes/report.ts
```diff
@@ -230,16 +222,41 @@ export default function (router: IRouter) {
...
-  if (error.response) {
-    try {
-      const esErrorResponse = JSON.parse(error.response);
-      return esErrorResponse.reason || error.response;
-    } catch (parsingError) {
-      return error.response;
+  // Delete single report by id
+  router.delete(
+    {
+      path: `${API_PREFIX}/reports/{reportId}`,
```
davidcui-amzn, line 229: Think this route should be something like `${API_PREFIX}/delete_report_by_id/{reportId}`

server/routes/reportDefinition.ts
```diff
@@ -0,0 +1,340 @@
...
+  REPORT_DEF_STATUS,
+  SCHEDULE_TYPE,
+  DELIVERY_CHANNEL,
+} from './utils/constants';
+
+export default function (router: IRouter) {
+  // Create report Definition
+  router.post(
+    {
+      path: `${API_PREFIX}/reportDefinition`,
```
davidcui-amzn, line 42: what do you think about renaming this to something like `createReportDefinition`? It looks a bit more descriptive to me

server/routes/reportDefinition.ts
```diff
@@ -0,0 +1,340 @@
...
+          body: parseEsErrorResponse(error),
+        });
+      }
+    }
+  );
+
+  // Update definition by id
+  router.put(
+    {
+      path: `${API_PREFIX}/reportDefinitions/{reportDefinitionId}`,
```
davidcui-amzn, line 106: same comment here: maybe a more descriptive route name?

server/routes/reportDefinition.ts
```diff
@@ -0,0 +1,340 @@
...
+          params
+        );
+
+        return response.ok({
+          body: esResp._id,
+        });
+      } catch (error) {
+        //@ts-ignore
+        context.reporting_plugin.logger.error(
+          `Fail to update report definition: ${error}`
```
davidcui-amzn, line 160: np: Failed vs Fail

server/routes/reportDefinition.ts
```diff
@@ -0,0 +1,340 @@
...
+          body: parseEsErrorResponse(error),
+        });
+      }
+    }
+  );
+
+  // get all report definitions details
+  router.get(
+    {
+      path: `${API_PREFIX}/reportDefinitions`,
```
davidcui-amzn, line 174: same comment here about descriptive routes

server/routes/report.ts
zhongnansu: yes, it extracts only the useful data for client side, gets rid of the redundant data fields returned by ES, such as `seq_num`, `version_num`, `id`

server/routes/reportDefinition.ts
```diff
@@ -0,0 +1,340 @@
...
+    }
+  );
+}
+
+function validateReportDefinition(reportDefinition: any) {
+  reportSchema.validate(reportDefinition);
+
+  const delivery = reportDefinition.delivery;
+  const trigger = reportDefinition.trigger;
+  const delivery_params = delivery.delivery_params;
```
davidcui-amzn, line 305: np: camel case

server/routes/report.ts
zhongnansu: https://restfulapi.net/resource-naming/
> URIs should not be used to indicate that a CRUD function is performed. URIs should be used to uniquely identify resources and not any action upon them. HTTP request methods should be used to indicate which CRUD function is performed.

server/routes/reportDefinition.ts
zhongnansu: uri naming shouldn't be descriptive of actions.

davidcui-amzn: Gotcha, TIL

server/routes/report.ts
davidcui-amzn: TIL

server/routes/reportDefinition.ts
zhongnansu: fixed

server/routes/reportDefinition.ts
zhongnansu: fixed, thx

server/model/index.ts
```diff
@@ -0,0 +1,118 @@
...
+  url: schema.uri(),
+  window_width: schema.number({ defaultValue: 1200 }),
+  window_height: schema.number({ defaultValue: 800 }),
+  report_format: schema.oneOf([schema.literal('pdf'), schema.literal('png')]),
+});
+
+export const scheduleSchema = schema.object({
+  schedule_type: schema.oneOf([
+    schema.literal('Now'),
+    schema.literal('Future Date'),
```
davidcui-amzn, line 35: Sorry i missed these but "Date" in Future date and "Based" in Cron based are not capitalized either

server/plugin.ts
```diff
@@ -60,6 +61,7 @@ export class OpendistroKibanaReportsPlugin
 
     // Register server side APIs
     registerReportRoutes(router);
+    registerReportDefinitionRoutes(router);
```
seraphjiang, line 64: try merge into one registerRoutes(router)
so we don't need to touch plugin.ts file when we introduce new router.

server/routes/reportDefinition.ts
```diff
@@ -0,0 +1,340 @@
...
+          sortField: schema.maybe(schema.string()),
+          sortDirection: schema.maybe(schema.string()),
+        }),
+      },
+    },
+    async (
+      context,
+      request,
+      response
+    ): Promise<IKibanaResponse<any | ResponseError>> => {
```
seraphjiang, line 187: anonymous method  is not friendly, in callstack when exception.

server/plugin.ts
zhongnansu: Make sense, fixed

PR #56, opened 2020-08-10
Build Reports Scheduler ES plugin

zhongnansu:
*Issue #, if available:*
#8 

*Description of changes:*
Build a ES plugin which extends the job-scheduler SPI. It provides APIs to schedule a report and apply locks to make sure when Kibana plugin is polling for jobs to execute, there is only one job get executed by one Kibana instance at a time.

- Set up reports-scheduler ES plugin and enable `gradle build`
- Restructure the project layout to have sub-projects in different folder
- Enable and expose the following APIs
  - create/delete scheduled job
  - get available job (acquire lock)
  - Update job status (release lock, remove job from queue)

passed `gradle build` and `gradle test` (only 1 IT test for now)


.gitignore
```diff
@@ -1,11 +1 @@
-npm-debug.log*
-node_modules
-/build/
-/public/app.css
-.idea/
-yarn-error.log
-/coverage/
 .DS_Store
```
akbhatta, line 1: I think it is better to have all required items in .gitignore

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
```diff
@@ -0,0 +1,221 @@
...
+                        } else {
+                            log.error(e.toString());
+                        }
+                    } else {
+                        log.error(e.toString());
+                    }
+                }
+            });
+
+//            IndexResponse response = client.prepareIndex("twitter", "_doc", "1")
```
akbhatta: remove commented code

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
```diff
@@ -0,0 +1,221 @@
...
+                public void onResponse(UpdateResponse updateResponse) {
+                    log.info("Existing job triggered and pushed to queue, waiting to be picked up by Kibana");
+                }
+
+                @Override
+                public void onFailure(Exception e) {
+                    if (e instanceof ElasticsearchException) {
+                        if (((ElasticsearchException) e).status() == RestStatus.NOT_FOUND) {
+                            jsonMap.put("report_definition_id", reportDefId);
+                            jsonMap.put("is_locked", false);
```
akbhatta: how is is_locked value synchronized with the lock service? last_locked_time makes more sense to handle timeout cases.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
```diff
@@ -0,0 +1,221 @@
...
+            throw new IllegalStateException("ClusterService is not initialized.");
+        }
+
+        if (this.threadPool == null) {
+            throw new IllegalStateException("ThreadPool is not initialized.");
+        }
+
+        Runnable runnable = () -> {
+
+            JobParameter parameter = (JobParameter) jobParameter;
```
akbhatta: nit: finals for the values (please check everywhere)

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
```diff
@@ -0,0 +1,221 @@
...
+            String reportDefId = parameter.getReportDefinitionId();
+
+
+            Map<String, Object> jsonMap = new HashMap<>();
+            jsonMap.put("to_be_executed", true);
```
akbhatta: what is the use of this value?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
```diff
@@ -0,0 +1,221 @@
...
+
+                            client.index(indexRequest, new ActionListener<IndexResponse>() {
+                                @Override
+                                public void onResponse(IndexResponse indexResponse) {
+                                    log.info("New job created and push to queue, waiting to be picked up by Kibana"); // TODO: better log message
+                                }
+
+                                @Override
+                                public void onFailure(Exception e) {
+                                    log.error(e.toString());
```
akbhatta: If small description is added (different per condition) then when debugging, it would be easy to identify condition.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
```diff
@@ -0,0 +1,148 @@
...
+import java.io.IOException;
+import java.time.Instant;
+
+/**
+ * A sample job parameter.
+ * <p>
+ * It adds an additional "indexToWatch" field to {@link ScheduledJobParameter}, which stores the index
+ * the job runner will watch.
+ */
+public class JobParameter implements ScheduledJobParameter {
```
akbhatta, line 32: why is this mutable? is it possible to get rid of setters?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/ReportsSchedulerRestHandler.java
```diff
@@ -0,0 +1,161 @@
...
+             *             "start_time": 1553112384
+             *         }
+             *
+             *       },
+             *     "report_definition_id":"23dasdfdasdas"
+             *  }
+             */
+
+            if(id == null) {
+                throw new IllegalArgumentException("Must specify id");
```
akbhatta: is it possible to handle this gracefully?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/ReportsSchedulerRestHandler.java
```diff
@@ -0,0 +1,161 @@
...
+                        }
+                    }
+
+                    @Override
+                    public void onFailure(Exception e) {
+                        restChannel.sendResponse(new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
+                    }
+                });
+            };
+        } else if (request.method().equals(RestRequest.Method.DELETE)) {
```
akbhatta: may be making use of POST/DELETE handlers instead of if/else would help unit test easily?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/ReportsSchedulerRestHandler.java
```diff
@@ -0,0 +1,161 @@
...
+                // index the job parameter
+                client.index(indexRequest, new ActionListener<IndexResponse>() {
+                    @Override
+                    public void onResponse(IndexResponse indexResponse) {
+                        try {
+                            RestResponse restResponse = new BytesRestResponse(RestStatus.OK,
+                                    indexResponse.toXContent(JsonXContent.contentBuilder(), null));
+                            restChannel.sendResponse(restResponse);
+                        } catch(IOException e) {
+                            restChannel.sendResponse(new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
```
akbhatta: is it always internal server error if not success!?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/ReportsSchedulerRestHandler.java
```diff
@@ -0,0 +1,161 @@
...
+            String id = request.param("id");
+            DeleteRequest deleteRequest = new DeleteRequest()
+                    .index(ReportsSchedulerPlugin.JOB_INDEX_NAME)
+                    .id(id);
+
+            return restChannel -> {
+                client.delete(deleteRequest, new ActionListener<DeleteResponse>() {
+                    @Override
+                    public void onResponse(DeleteResponse deleteResponse) {
+                        restChannel.sendResponse(new BytesRestResponse(RestStatus.OK, "Job deleted."));
```
akbhatta: what does "Job deleted." means to caller?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
```diff
@@ -0,0 +1,134 @@
...
+     */
+    public ReportsJobActionHandler(NodeClient client, RestChannel channel, ClusterService clusterService) {
+        super(client, channel);
+        this.lockService = new LockService(client, clusterService);
+    }
+
+
+    public void getJob() {
+        // get all jobs from queue_index
+        // return 404 if no jobs in the queue, or the queue doesn't exist at all
```
akbhatta: 404 may not be right error. check 204 if it makes sense.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
```diff
@@ -0,0 +1,134 @@
...
+        SearchRequest searchRequest = new SearchRequest(ReportsSchedulerJobRunner.JOB_QUEUE_INDEX_NAME);
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder();
+        searchSourceBuilder.query(QueryBuilders.matchAllQuery());
+
+        // sort the document
+        searchSourceBuilder.sort(new FieldSortBuilder("enqueue_time").order(SortOrder.DESC));
```
akbhatta: sorting may lead to clashing. randomizing would reduce clashing and faster access of jobs (instead of serial waiting for locks)

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
```diff
@@ -0,0 +1,134 @@
...
+        SearchHits hits = response.getHits();
+        SearchHit[] searchHits = hits.getHits();
+
+        // loop to retrieve job that is not locked, and then acquire lock
+        for (SearchHit hit : searchHits) {
+            String jobId = hit.getId();
+            // TODO: duplicate code block
+            JobParameter jobParameter = new JobParameter();
+            jobParameter.setLockDurationSeconds(LOCK_DURATION_SECONDS);
+            JobExecutionContext ctx = new JobExecutionContext(null, null, null, ReportsSchedulerJobRunner.JOB_QUEUE_INDEX_NAME, jobId);
```
akbhatta: descriptive variable names would be easy to read code (other places as well).

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
zhongnansu: please review the latest changes from all commits

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
zhongnansu: You've reviewing the past commits, that's removed in later commits.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
zhongnansu: That's removed in later commits.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
zhongnansu: removed in later commits, mark as resolved

reports-scheduler/NOTICE
```diff
@@ -0,0 +1,2 @@
+Open Distro for Elasticsearch Job Scheduler
+Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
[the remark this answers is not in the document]
zhongnansu: good catch.

.gitignore
zhongnansu: good point, will do

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/ReportsSchedulerRestHandler.java
zhongnansu: In the latest change, it's returning something like ""Job deleted. Job id: <jobID>". Does that make sense? 
reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsScheduleActionHandler.java

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/ReportsSchedulerRestHandler.java
zhongnansu: Yes, the action is save data to elasticsearch index, by using the client

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
zhongnansu: changed to 204 in later commits. Resolve this

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
zhongnansu: sure, will change

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
zhongnansu: Got your point. Added

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
zhongnansu: will do

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
zhongnansu: made variables final and removed setters

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
```diff
@@ -0,0 +1,139 @@
...
+      INSTANCE = new ReportsSchedulerJobRunner();
+      return INSTANCE;
+    }
+  }
+
+  private ReportsSchedulerJobRunner() {
+    // Singleton class, use getJobRunner method instead of constructor
+  }
+
+  public void setClient(Client client) {
```
akbhatta, line 79: Why is this class mutable (having setters)?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
```diff
@@ -0,0 +1,136 @@
...
+  @Override
+  public Long getLockDurationSeconds() {
+    return this.lockDurationSeconds;
+  }
+
+  public String getReportDefinitionId() {
+    return this.reportDefinitionId;
+  }
+
+  //  public void setJobName(String jobName) {
```
akbhatta, line 91: Please remove commented code.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
```diff
@@ -0,0 +1,136 @@
...
+  //
+  //  public void setSchedule(Schedule schedule) {
+  //    this.schedule = schedule;
+  //  }
+  //
+  //  public void setReportDefinitionId(String reportDefinitionId) {
+  //    this.reportDefinitionId = reportDefinitionId;
+  //  }
+
+  public void setLockDurationSeconds(Long lockDurationSeconds) {
```
akbhatta, line 115: Why is this class mutable?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsScheduleAction.java
```diff
@@ -0,0 +1,80 @@
...
+import org.elasticsearch.rest.RestRequest;
+import org.elasticsearch.rest.RestStatus;
+
+import com.amazon.opendistroforelasticsearch.reportsscheduler.rest.handler.ReportsScheduleActionHandler;
+import com.google.common.collect.ImmutableList;
+
+public class RestReportsScheduleAction extends BaseRestHandler {
+  public static final String SCHEDULER_SCHEDULE_ACTION = "reports_scheduler_schedule_action";
+  private final Settings settings;
+  private final String SCHEDULE = "schedule";
```
akbhatta, line 36: "static" is missing

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsJobAction.java
```diff
@@ -0,0 +1,81 @@
...
+import org.elasticsearch.rest.BytesRestResponse;
+import org.elasticsearch.rest.RestRequest;
+import org.elasticsearch.rest.RestStatus;
+
+import com.amazon.opendistroforelasticsearch.reportsscheduler.rest.handler.ReportsJobActionHandler;
+import com.google.common.collect.ImmutableList;
+
+public class RestReportsJobAction extends BaseRestHandler {
+  public static final String SCHEDULER_JOB_ACTION = "reports_scheduler_job_action";
+  private final String JOB = "job";
```
akbhatta, line 36: "static" is missing

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsJobAction.java
```diff
@@ -0,0 +1,81 @@
...
+  public String getName() {
+    return SCHEDULER_JOB_ACTION;
+  }
+
+  @Override
+  public List<Route> routes() {
+    return ImmutableList.of(
+        new Route(
+            RestRequest.Method.POST,
+            String.format(Locale.ROOT, "%s/%s/{%s}", BASE_SCHEDULER_URI, JOB, JOB_ID)),
```
akbhatta, line 57: direct concatenation is better than using String.format (both for CPU and Memory)

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsScheduleAction.java
```diff
@@ -0,0 +1,80 @@
...
+  public String getName() {
+    return SCHEDULER_SCHEDULE_ACTION;
+  }
+
+  @Override
+  public List<Route> routes() {
+    return ImmutableList.of(
+        new Route(
+            RestRequest.Method.POST,
+            String.format(Locale.ROOT, "%s/%s", BASE_SCHEDULER_URI, SCHEDULE)),
```
akbhatta, line 52: direct concatenation is better than using String.format (both for CPU and Memory)

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsScheduleAction.java
```diff
@@ -0,0 +1,80 @@
...
+        new Route(
+            RestRequest.Method.DELETE,
+            String.format(Locale.ROOT, "%s/%s", BASE_SCHEDULER_URI, SCHEDULE)));
+  }
+
+  @Override
+  protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+    final String jobId = request.param("job_id");
```
akbhatta, line 60: probably "job_id" can be extracted to constant

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/AbstractActionHandler.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.elasticsearch.client.node.NodeClient;
+import org.elasticsearch.rest.BytesRestResponse;
+import org.elasticsearch.rest.RestChannel;
+
+/** Action handler to process REST request and handle failures. */
+public abstract class AbstractActionHandler {
+  protected final NodeClient client;
```
akbhatta, line 28: why protected instead of getters?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
```diff
@@ -0,0 +1,192 @@
...
+   *
+   * @param searchHitsQueue a queue which saves all jobs from .reports_scheduler_job_queue index
+   */
+  private void findFirstAvailableJob(Queue<SearchHit> searchHitsQueue) {
+    final SearchHit hit = searchHitsQueue.poll();
+    log.info("queue size after poll: " + searchHitsQueue.size());
+
+    if (hit != null) {
+      String jobId = hit.getId();
+      // set up jobParamater and jobContext that is required by lockService
```
akbhatta, line 109: is this TODO: ?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
```diff
@@ -0,0 +1,192 @@
...
+          jobParameter,
+          jobExecutionContext,
+          ActionListener.wrap(
+              lock -> {
+                if (lock == null) {
+                  findFirstAvailableJob(searchHitsQueue);
+                } else {
+                  log.info("send job data(report_definition_id) to reporting core for execution");
+
+                  RestResponse restResponse =
```
akbhatta, line 125: final missing.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
```diff
@@ -0,0 +1,192 @@
...
+                      new BytesRestResponse(
+                          RestStatus.OK, hit.toXContent(JsonXContent.contentBuilder(), null));
+                  channel.sendResponse(restResponse);
+                }
+              },
+              exception -> {
+                channel.sendResponse(
+                    new BytesRestResponse(
+                        RestStatus.INTERNAL_SERVER_ERROR, "Failed to acquire lock"));
+                throw new IllegalStateException("Failed to acquire lock.");
```
akbhatta, line 135: Not sure if throwing exception is the right thing here

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsScheduleActionHandler.java
```diff
@@ -0,0 +1,109 @@
...
+            .source(request.requiredContent(), XContentType.JSON);
+
+    // index the job parameter
+    client.index(
+        indexRequest,
+        new ActionListener<IndexResponse>() {
+          @Override
+          public void onResponse(IndexResponse indexResponse) {
+            try {
+              RestResponse restResponse =
```
akbhatta, line 63: final missing

reports-scheduler/src/test/java/com/amazon/opendistroforelasticsearch/reportsscheduler/ReportsSchedulerPluginIT.java
```diff
@@ -0,0 +1,51 @@
...
+
+public class ReportsSchedulerPluginIT extends ESIntegTestCase {
+
+  public void testPluginsAreInstalled() {
+    ClusterHealthRequest request = new ClusterHealthRequest();
+    ClusterHealthResponse response =
+        ESIntegTestCase.client().admin().cluster().health(request).actionGet();
+    Assert.assertEquals(ClusterHealthStatus.GREEN, response.getStatus());
+
+    NodesInfoRequest nodesInfoRequest = new NodesInfoRequest();
```
akbhatta, line 38: final missing

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsScheduleActionHandler.java
```diff
@@ -0,0 +1,109 @@
...
+                  new BytesRestResponse(
+                      RestStatus.OK,
+                      String.format(Locale.ROOT, "Job de-scheduled. Job id: %s", jobId)));
+            }
+          }
+
+          @Override
+          public void onFailure(Exception e) {
+            channel.sendResponse(
+                new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
```
akbhatta, line 105: Why is this always internal server error?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsScheduleActionHandler.java
```diff
@@ -0,0 +1,109 @@
...
+              channel.sendResponse(restResponse);
+            } catch (IOException e) {
+              channel.sendResponse(
+                  new BytesRestResponse(RestStatus.INTERNAL_SERVER_ERROR, e.getMessage()));
+            }
+          }
+
+          @Override
+          public void onFailure(Exception e) {
+            channel.sendResponse(
```
akbhatta, line 75: why is this always internal server error?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
```diff
@@ -0,0 +1,192 @@
...
+                                    "Job deleted from jobs queue. Job id: %s",
+                                    jobId)));
+                      },
+                      exception -> log.debug("Failed to delete lock", exception)));
+            }
+          }
+
+          @Override
+          public void onFailure(Exception e) {
+            channel.sendResponse(
```
akbhatta, line 187: why is this always internal server error?

reports-scheduler/src/test/java/com/amazon/opendistroforelasticsearch/reportsscheduler/ReportsSchedulerPluginIT.java
zhongnansu: added

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
zhongnansu: No, this is the comment for the following lines, which creates a "fake" jobparamater and jobcontext instance, required by initializing lockService instance

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsScheduleAction.java
zhongnansu: good point, fixed

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
zhongnansu: Because it doesn't need locking when we created the scheduled job, but it needs locking when we push the job to queue. That's why I left this variable with a setter. But I can add locking for scheduled job too, I just feel like there is no need for that. 

In job-scheduler document it says `ScheduledJob executions are not a singleton runner and if a job execution takes longer than the interval/cron time then there can be two or more execution running at a time.`

In our use case, the "job execution" is nothing complex, just pushing scheduled job to queue index, it's just a "save to index" action. And it won't take longer than the interval time we set.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsJobAction.java
zhongnansu: Do you mean something like this, `new Route(RestRequest.Method.POST, BASE_SCHEDULER_URI + "/" + JOB + "/" + "{" + JOB_ID + "}")`, I'll say the readability is really bad

zhongnansu: Besides, I checked other plugin repos in Open Distro. They all defined it the same way. e.g. https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/92a586718f93a9ab934169cab0aca42c8641f4d9/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/RestAnomalyDetectorJobAction.java#L103

akbhatta: May be extract the url formation to constants `private static final String JOB_ID_URL = BASE_SCHEDULER_URI + "/" + JOB + "/{" + JOB_ID + "}";` and use it in function call.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
zhongnansu: They are all failure response from operations on database(delete or create). What do you suggest? (for the same comments you made else where)

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/AbstractActionHandler.java
zhongnansu: Only allowing subclass to access this variable. I think protected is the first thing come to my mind
https://github.com/opendistro-for-elasticsearch/anomaly-detection/blob/92a586718f93a9ab934169cab0aca42c8641f4d9/src/main/java/com/amazon/opendistroforelasticsearch/ad/rest/handler/AbstractActionHandler.java#L31

akbhatta: would protected getter method help here?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
akbhatta: more appropriate HTTP error codes depending on the root cause of the issue? may be identify the root cause depending on "e"? say the lock is already released/deleted and second call fails then return "Resource not found".

zhongnansu: The scenario you mentioned is handled here. https://github.com/opendistro-for-elasticsearch/kibana-reports/blob/ff733205fe41408133bb6b3e85e5752bb5a50d95/reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java#L164

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/AbstractActionHandler.java
zhongnansu: protected getter should work, could you explain why protected getter is preferred?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsJobAction.java
zhongnansu: fixed

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
zhongnansu: changed it to `log.debug("Fail to acquire lock " + exception)`

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
zhongnansu: The job runner should be a singleton class if it uses Elasticsearch client or other objects passed from Elasticsearch. Because when registering the job runner to JobScheduler plugin, Elasticsearch has not invoke plugins' createComponents() method. That is saying the plugin is notcompletely initialized, and the Elasticsearch client and clusterService and other objects are not available to plugin and this job runner.

So we have to move this job runner initialization to `createComponents()` method in reportsSchedulerPlugin, and using singleton job runner to ensure we register a usable job runner instance to JobScheduler plugin.

Tried to declare it as final, and create the instance using constructor in `createdComponents()`and there is always exception thrown saying `clusterService has not been initialized`

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
```diff
@@ -0,0 +1,112 @@
...
+  public XContentBuilder toXContent(XContentBuilder builder, Params params) throws IOException {
+    builder.startObject();
+    builder
+        .field(JobConstant.NAME_FIELD, this.jobName)
+        .field(JobConstant.ENABLED_FILED, this.isEnabled)
+        .field(JobConstant.SCHEDULE_FIELD, this.schedule)
+        .field(JobConstant.REPORT_DEFINITION_ID, this.reportDefinitionId);
+    if (this.enabledTime != null) {
+      builder.timeField(
+          JobConstant.ENABLED_TIME_FILED,
```
davidcui-amzn, line 105: np: consistent indentation

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsJobAction.java
```diff
@@ -0,0 +1,78 @@
...
+import org.elasticsearch.rest.BaseRestHandler;
+import org.elasticsearch.rest.BytesRestResponse;
+import org.elasticsearch.rest.RestRequest;
+import org.elasticsearch.rest.RestStatus;
+
+import com.amazon.opendistroforelasticsearch.reportsscheduler.rest.handler.ReportsJobActionHandler;
+import com.google.common.collect.ImmutableList;
+
+public class RestReportsJobAction extends BaseRestHandler {
+  public static final String SCHEDULER_JOB_ACTION = "reports_scheduler_job_action";
```
davidcui-amzn, line 34: np: consistency with camel case or snake case

reports-scheduler/src/test/java/com/amazon/opendistroforelasticsearch/reportsscheduler/ReportsSchedulerPluginIT.java
```diff
@@ -0,0 +1,51 @@
...
+import org.elasticsearch.action.admin.cluster.node.info.NodesInfoResponse;
+import org.elasticsearch.action.admin.cluster.node.info.PluginsAndModules;
+import org.elasticsearch.cluster.health.ClusterHealthStatus;
+import org.elasticsearch.plugins.PluginInfo;
+import org.elasticsearch.test.ESIntegTestCase;
+import org.junit.Assert;
+
+public class ReportsSchedulerPluginIT extends ESIntegTestCase {
+
+  public void testPluginsAreInstalled() {
```
davidcui-amzn, line 32: Will we add any correctness tests? Seems like an incomplete test suite if it's only testing that plugins are installed

zhongnansu: Tests will be added later, in another PR.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
zhongnansu: I am using google-java-format to format the code, the same as Elasticsearch source code and ODFE sql. It's using 2-space indentation. It seems like when you start a new line but in the same statement, it uses 4-space
https://github.com/opendistro-for-elasticsearch/sql/blob/2939081a1927f8886cd9c16cb03430b6a6352c6e/ppl/src/main/java/com/amazon/opendistroforelasticsearch/sql/ppl/PPLService.java#L60

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsJobAction.java
zhongnansu: It's literally snake_case, it's the naming convention of constants.
> The names of variables declared class constants and of ANSI constants should be all uppercase with words separated by underscores ("_"). (ANSI constants should be avoided, for ease of debugging.

davidcui-amzn: Sorry, I was referring to the `"reports_scheduler_job_action"` not the variable name

reports-scheduler/src/test/java/com/amazon/opendistroforelasticsearch/reportsscheduler/ReportsSchedulerPluginIT.java
davidcui-amzn: 👍

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/RestReportsJobAction.java
zhongnansu: changed "jobId" to "job_id" to use snake case in url 
https://github.com/elastic/elasticsearch/blob/20c4da3a64e3b1ccf5cef26e9e99b561df4efa76/plugins/examples/rest-handler/src/main/java/org/elasticsearch/example/resthandler/ExampleCatAction.java#L47

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/ReportsSchedulerJobRunner.java
akbhatta: May be in that case I think that better way is to not create instance in getJobRunnerInstance() but only return if it exists. createComponents() will create object with proper initialization. anyone calling getJobRunnerInstance() before that will get InvalidState exception. with that we can assure immutability?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
akbhatta: With that information, the appropriate object can be created with proper value during construction to avoid mutability.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/AbstractActionHandler.java
akbhatta: OOP principle of encapsulation.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
akbhatta: in what all scenarios onFailure() is expected?

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/AbstractActionHandler.java
zhongnansu: will make variables private final, and use protected getters.

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/job/parameter/JobParameter.java
zhongnansu: will make it final and use constructors for initialization, like other variables

reports-scheduler/src/main/java/com/amazon/opendistroforelasticsearch/reportsscheduler/rest/handler/ReportsJobActionHandler.java
zhongnansu: found some reference here about the Elasticsearch high-level-client exceptions, which is exactly what we are using here.
issue: https://github.com/elastic/elasticsearch/issues/30334
javaDoc: https://github.com/elastic/elasticsearch/blob/f56ab039aec803c22621b82bd4ea30441dd19b8c/client/rest-high-level/src/main/java/org/elasticsearch/client/RestHighLevelClient.java#L240
> an {@link IOException} is usually thrown in case of failing to parse the REST response in the high-level REST client, the request times out or similar cases where there is no response coming back from the Elasticsearch server
an {@link ElasticsearchException} is usually thrown in case where the server returns a 4xx or 5xx error code. The high-level client then tries to parse the response body error details into a generic ElasticsearchException and suppresses the original {@link ResponseException}

Two types of exceptions are thrown by the `client`. My idea is to assign status code 502 to IOException and 503 to ElasticsearchExeception. What do you think?

> 502 Bad Gateway
The server was acting as a gateway or proxy and received an invalid response from the upstream server.[65]
503 Service Unavailable
The server cannot handle the request (because it is overloaded or down for maintenance). Generally, this is a temporary state.[66]
504 Gateway Timeout
The server was acting as a gateway or proxy and did not receive a timely response from the upstream server.[67]

akbhatta: looks good

zhongnansu:
> nit: final

fixed

Issue #98, opened 2020-09-25
Access kibana.yml from plugin to read server.host, server.port and server.basePath


zhongnansu:
Currently those settings are not accessible from new Kibana plugin platform. Only available settings are listed here. https://github.com/elastic/kibana/blob/faf4b040042918fe804bc3f80e42011a073c2d8c/src/core/server/plugins/types.ts#L262

There are also issues in the community
- https://discuss.elastic.co/t/kibana-7-6-1-getting-values-from-kibana-yml/227076/6
- https://discuss.elastic.co/t/unknown-key-elasticsearch-hosts-for-kibana-plugin-config/211254
- https://discuss.elastic.co/t/kibana-plugin-read-configuration-from-kibana-yml/122260
- https://discuss.elastic.co/t/access-variables-in-kibana-yml/226874

Kibana Code reference for deep dive RCA
- https://github.com/elastic/kibana/blob/3759063e813f364c39f2fd93d59827319a4681a5/src/cli/cluster/cluster_manager.ts#L151 This plugin is accessing server.host from kibana.yml 
- https://github.com/elastic/kibana/blob/faf4b040042918fe804bc3f80e42011a073c2d8c/src/core/server/legacy/types.ts#L42 Legacy config interface

There is not a proper way to fix this now, we'll wait for the official support from NP in the future.

zhongnansu:
Seems like there is a fix for the NP in upstream elastic/kibana. At least we can access `server.basePath`
https://github.com/elastic/kibana/issues/45991

doc ref: https://github.com/elastic/kibana/blob/v7.10.0/docs/development/core/server/kibana-plugin-core-server.basepath.md

zhongnansu:
A possible solution here, we are looking into it
https://github.com/opendistro-for-elasticsearch/kibana-reports/pull/245#issuecomment-747790082

zhongnansu:
Ref: https://www.elastic.co/guide/en/kibana/current/reporting-settings-kb.html#reporting-kibana-server-settings

PR #99, opened 2020-09-25
Use kibana server hostname for puppeteer to access pages

zhongnansu:
*Issue #, if available:*
Currently the url for puppeteer to access is a complete url passed from UI, including the schema, hostname and port. However this brings some issues when deploying on a cloud service such as AWS ELB, which has settings for security group. Since puppeteer is only accessing the url from localhost, it doesn't even need to go the ELB first and route back to itself. Instead it can access `http://localhost:5601:/app/dashboard#/<dashboard_id>` directly.

*Description of changes:*
- This is a temporary fix to replace url prefix with a hardcoded `LOCAL_HOST = "http://localhost:5601"` constant. However since the hostname and port number are configured in `kibana.yml` file, which the new Kibana plugin platform doesn't provide access now. This issue is documented #98  and tracked once there is a proper solution.
- remove unused dependencies, move some packages from `dependency` to `devDependency`


kibana-reports/server/routes/utils/constants.ts
```diff
@@ -66,3 +66,5 @@ export enum TRIGGER_TYPE {
 }
 // https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-from-size.html
 export const DEFAULT_MAX_SIZE = 10000;
+
+export const LOCAL_HOST = 'http://localhost:5601';
```
davidcui-amzn, line 70: Should we be hard-coding the host like this?

kibana-reports/server/routes/utils/dataReportHelpers.ts
```diff
@@ -1,4 +1,3 @@
-import { async } from 'rxjs/internal/scheduler/async';
```
davidcui-amzn, line 1: Are the changes in this file related to the title of the PR? I feel this is a different optimization refactor

zhongnansu: That's just an unused import

kibana-reports/server/routes/utils/constants.ts
zhongnansu: Ideally, we should extract the server host/port from `kibana.yml`, but the new platform applies some restrictions to access certain fields. More details can be found in this issue #98

kibana-reports/server/routes/utils/dataReportHelpers.ts
davidcui-amzn: Not just this line, I'm referring to the other changes in the file as well- meant to highlight the whole file

kibana-reports/server/routes/utils/constants.ts
davidcui-amzn: Can we get the local host from a `window.location` or `document` method instead? I think those should be fine to access in the new platform.

zhongnansu: That `window.location` won't work on server side. The UI side still keeps the usage of `window.location` which is fine.

davidcui-amzn: Ack, miss from me- forgot this was on server-side

davidcui-amzn (approved):
LGTM- just some small comments and questions to address

kibana-reports/server/routes/utils/dataReportHelpers.ts
zhongnansu: The other change is just rewriting the import to typescript style.

akbhatta (approved):

Issue #40, opened 2020-07-10
plugin fails to run after installed to kibana


zhongnansu:
**Issue Description**

After `yarn build` generate the distribution `.zip` file, I tried to install to `kibana` by

```
./bin/kibana-plugin install file:///Users/szhongna/Desktop/reporting/kibana/plugins/kibana-reports/build/opendistro_kibana_reports-0.0.1.zip
``` 

However, when I run `./bin/kibana`, it throws the following error.

**error log**
```
  log   [23:43:59.931] [warning][plugins-discovery] Expect plugin "id" in camelCase, but found: apm_oss
  log   [23:44:00.891] [info][plugins-service] Plugin "visTypeXy" is disabled.
  log   [23:44:00.926] [fatal][root] { Error: Cannot find module '/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/plugins/opendistro_kibana_reports'
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
    at Module.Hook._require.Module.require (/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/node_modules/require-in-the-middle/index.js:51:29)
    at require (internal/modules/cjs/helpers.js:25:18)
    at createPack (/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/src/legacy/plugin_discovery/plugin_pack/create_pack.js:33:18)
    at MapSubscriber.packageJson$.pipe [as project] (/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/src/legacy/plugin_discovery/plugin_pack/create_pack.js:66:11)
    at MapSubscriber._next (/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/node_modules/rxjs/internal/operators/map.js:49:35)
    at MapSubscriber.Subscriber.next (/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/node_modules/rxjs/internal/Subscriber.js:66:18)
    at RefCountSubscriber.Subscriber._next (/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/node_modules/rxjs/internal/Subscriber.js:89:26)
    at RefCountSubscriber.Subscriber.next (/Users/szhongna/Downloads/kibana-7.8.0-darwin-x86_64/node_modules/rxjs/internal/Subscriber.js:66:18)
[...]

PR #101, opened 2020-09-29
Upgrade to ES/Kibana version v7.9.1

zhongnansu:
*Issue #, if available:*

*Description of changes:*
Upgrade to ES/Kibana version v7.9.1 for both reports-scheduler plugin and kibana-reports plugin
- Kibana reports plugin
  - server
    - replace `IClusterClient` ,`IScopedClusterClient` to`ILegacyClusterClient`, `ILegacyScopedClusterClient`
    - replace `...dataClient.callAsCurrentUser` to `...legacy.client.callAsCurrentUser`
  - UI 
    - add `data` to `kibana.json` -> `requiredPlugins`
- Reports scheduler
  - bump version/dependency number

- This update also solves one of the kibana plugin build issue in #40. The issue that  `kibana-plugin-helper` didn't pick up hidden files started with dot, which in our case is the `.local-chromium`


akbhatta (approved):

kibana-reports/server/plugin.ts
```diff
@@ -19,7 +19,7 @@ import {
   CoreStart,
   Plugin,
   Logger,
-  IClusterClient,
+  ILegacyClusterClient,
```
davidcui-amzn, line 22: We revert back to legacy for 7.9?

davidcui-amzn (commented):
Does the UI still render with the version change when navigating to the Reporting homepage?

kibana-reports/server/plugin.ts
zhongnansu: yes, the `IClusterClient` in kibana v7.9.1 is not fully supported for our usage.

zhongnansu:
> Does the UI still render with the version change when navigating to the Reporting homepage?

The only issues that UI is complaining is the now it requires a `data` plugins in the kibana.json. Everything looks good after I added that

kibana-reports/server/plugin.ts
davidcui-amzn: Got it

davidcui-amzn (approved):
LGTM

PR #110, opened 2020-10-07
Fix enable/disable after editing schedule type

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
Enabling and disabling a schedule report definition is fixed when the definition is edited between Recurring and Cron based.

(e.g if I create a `Recurring` definition and then edit it to become a `Cron based` definition, I can now `enable` and `disable` that updated scheduled definition without an error). 


kibana-reports/public/components/report_definitions/report_trigger/report_trigger.tsx
```diff
@@ -95,11 +95,16 @@ export function ReportTrigger(props: ReportTriggerProps) {
     reportDefinitionRequest.trigger.trigger_type = e;
     if (e === 'On demand') {
       delete reportDefinitionRequest.trigger.trigger_params;
-    } 
+    }
   };
 
   const handleScheduleType = (e: React.SetStateAction<string>) => {
     setScheduleType(e);
+    if (e === 'Cron based') {
```
akbhatta, line 103: When same string is used for comparison in multiple places, may be it is better to have it as named constant?

akbhatta (approved):

anirudha (approved):

kibana-reports/public/components/report_definitions/report_trigger/report_trigger.tsx
davidcui-amzn: Will change

PR #117, opened 2020-10-08
Context Menu Fixes 

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
Fixed some bugs in the context menu.
* The link in the `Success` toast to the `Reports` homepage works now
* Pre-selection of a report source after navigating to `Create report definition` from the context menu works
* Toasts now disappear after `6 seconds` instead of requiring the user closing manually. 

### DIFF
--- a/kibana-reports/public/components/report_definitions/report_settings/report_settings.tsx
+++ b/kibana-reports/public/components/report_definitions/report_settings/report_settings.tsx
@@ -144,7 +144,7 @@ export function ReportSettings(props: ReportSettingProps) {
   }) => {
     setDashboardSourceSelect(e.target.value);
     reportDefinitionRequest.report_params.core_params.base_url =
-      getDashboardBaseUrlCreate() + e.target.value;
+      getDashboardBaseUrlCreate(edit, editDefinitionId) + e.target.value;
   };
 
   const handleVisualizationSelect = (e: {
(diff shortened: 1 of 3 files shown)


zhongnansu (approved):

akbhatta (approved):

kibana-reports/public/components/report_definitions/report_settings/report_settings.tsx
```diff
@@ -398,6 +398,18 @@ export function ReportSettings(props: ReportSettingProps) {
     }
   };
 
+  const setInContextDefaultConfiguration = () => {
+    const url = window.location.href;
+    const id = parseInContextUrl(url, 'id');
+    if (url.includes('dashboard')) {
```
joshuali925, line 404: This is a corner case, but if I have a visualization which title contains `dashboard`, would the source be identified as dashboard? and would that break Reporting?

joshuali925 (approved):
code looks good, one question

kibana-reports/public/components/report_definitions/report_settings/report_settings.tsx
zhongnansu: good catch

davidcui-amzn: No- the url constructed doesn't use the name of the report source but the id. `dashboard` or `visualization` is included in the url depending on which page we came from, a dashboard or url.

joshuali925: got it, thanks

PR #120, opened 2020-10-10
Header & Footer Plaintext Support 

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
Header and footer have plain-text support and will display in the report header and footer as plaintext.

`TODO`: Add formatting to the report to clean up the display of the header and footer with the report. 

### DIFF
--- a/kibana-reports/public/components/report_definitions/create/create_report_definition.tsx
+++ b/kibana-reports/public/components/report_definitions/create/create_report_definition.tsx
@@ -35,6 +35,8 @@ interface reportParamsType {
   report_name: string;
   report_source: string;
   description: string;
+  header: string;
+  footer: string;
   core_params: visualReportParams | dataReportParams;
 }
 interface visualReportParams {
@@ -98,6 +100,8 @@ export function CreateReport(props) {
       report_name: '',
       report_source: '',
       description: '',
+      header: '',
+      footer: '',
       core_params: {
         base_url: '',
         report_format: '',
--- a/kibana-reports/public/components/report_definitions/create/create_report_definition.tsx
+++ b/kibana-reports/public/components/report_definitions/create/create_report_definition.tsx
@@ -35,13 +35,13 @@ interface reportParamsType {
   report_name: string;
   report_source: string;
   description: string;
-  header: string;
-  footer: string;
   core_params: visualReportParams | dataReportParams;
 }
 interface visualReportParams {
   base_url: string;
   report_format: string;
+  header: string;
+  footer: string;
   time_duration: string;
 }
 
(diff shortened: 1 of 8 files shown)


joshuali925 (approved):

kibana-reports/public/components/report_definitions/create/create_report_definition.tsx
```diff
@@ -35,6 +35,8 @@ interface reportParamsType {
   report_name: string;
   report_source: string;
   description: string;
+  header: string;
+  footer: string;
```
zhongnansu, line 39: This is only for create report definition, also need to add support for edit report definition

kibana-reports/server/model/index.ts
```diff
@@ -113,7 +113,8 @@ export const reportDefinitionSchema = schema.object({
       schema.literal(REPORT_TYPE.savedSearch),
     ]),
     description: schema.string(),
-
+    header: schema.string(),
+    footer: schema.string(),
```
zhongnansu, line 117: If you add the fields here, which means header and footer are for all types of report. I don't think we need header/footer for csv report, can you confirm? If so, add header/footer fields inside the  `visualReportSchema`. If these fields are optional, use `header: schema.maybe(schema.string())`

zhongnansu (commented):
Please address the comments

kibana-reports/public/components/report_definitions/report_settings/report_settings.tsx
```diff
@@ -207,6 +207,16 @@ export function ReportSettings(props: ReportSettingProps) {
       'write' | 'preview'
     >('write');
 
+    const handleHeader = (e) => {
+      setHeader(e);
+      reportDefinitionRequest.report_params.header = e;
+    };
```
zhongnansu, line 213: Any input validation for header and footer? Can user choose not to add any header and footer at all?

zhongnansu:
I didn't see any code change in `report detail` or `report definition detail`. It may not be displayed on those two pages.

kibana-reports/public/components/report_definitions/create/create_report_definition.tsx
davidcui-amzn: Good catch, will add

kibana-reports/public/components/report_definitions/report_settings/report_settings.tsx
davidcui-amzn: Yes, header and footer can be blank- not sure what input validation to include, hasn't been discussed before.

zhongnansu (approved):

Issue #90, opened 2020-09-17
Improve input validation on client side and server side


zhongnansu:
Currently we have some basic input validation on the server side. 
Client also needs some input validation, server side needs to add more input validation

invalid input includes:
- [x] empty fields, e.g. no report name
- [x] invalid fields
- [ ] TODO

PR #128, opened 2020-10-19
Improve server side input validation 

zhongnansu:
*Issue #, if available:*
#90 

*Description of changes:*
- input validation
  - url: pass relative url instead of full url, compose url on server side wiht `localhost:5601`
  - text fields: validate non-empty.
     - `report Name`: validate allowed characters using regex
     -  `email recipients`: validate email address format
  - timestamp fields: validate positive number
  - cron-expression: validate unix cron-expression (5 digits)
  
  - time duration: validate ISO format for time duration. e.g. `PT10M` (10 mintues)
- Decouple create visual report functions to another file, prepare for project layout re-arrangement. Will implement in future PRs.

### DIFF
--- a/kibana-reports/public/components/main/main_utils.tsx
+++ b/kibana-reports/public/components/main/main_utils.tsx
@@ -63,7 +63,7 @@ export const addReportsTableContent = (data) => {
       //TODO: wrong name
       timeCreated: report.time_created,
       state: report.state,
-      url: report.query_url,
+      url: `${location.host}${report.query_url}`,
       format: reportParams.core_params.report_format,
     };
     reportsTableItems.push(reportsTableEntry);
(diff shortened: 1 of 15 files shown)


kibana-reports/public/components/main/main_utils.tsx
```diff
@@ -63,7 +63,7 @@ export const addReportsTableContent = (data) => {
       //TODO: wrong name
       timeCreated: report.time_created,
       state: report.state,
-      url: report.query_url,
+      url: `${location.host}${report.query_url}`,
```
davidcui-amzn, line 66: What does `location.host` add?

kibana-reports/server/model/index.ts
```diff
@@ -27,32 +29,56 @@ import {
 } from '../routes/utils/constants';
 
 export const dataReportSchema = schema.object({
-  base_url: schema.uri(),
+  base_url: schema.string({
+    validate(value) {
```
davidcui-amzn, line 33: If the validation fails will this still return an error to the client-side and trigger the error toasts in the UI?

davidcui-amzn (approved):
LGTM- just have some minor questions

kibana-reports/public/components/main/main_utils.tsx
zhongnansu: the `query_url` is now relative, need to add the host as well for the link rendered in report table -> report source field to be accessible.

kibana-reports/server/model/index.ts
zhongnansu: It will, it will also include the error message

akbhatta (approved):

Issue #88, opened 2020-09-17
auto-refresh report/report definition table after create a report definition is not working


zhongnansu:
auto-refresh report/report definition table after create a report definition is not working

zhongnansu:
Also while one report is under "pending status", we should disable all links(report detail, report source, download icon)

PR #139, opened 2020-10-22
Workaround to fix table refresh, disable links for pending reports

joshuali925:
*Issue #, if available:*
- #88 

*Description of changes:*
- refresh table twice on load to fetch newly added reports and report definitions (this is only a workaround, I don't see how to trigger `<Main />` to refresh from `<CreateReport />` component, let me know if there is a better implementation)
- disable clickable links in reports table for pending reports
image


davidcui-amzn (approved):
LGTM

zhongnansu (approved):

PR #144, opened 2020-10-23
Fix Timezone selection and bugs 

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
* Temporarily removed timezone from daily/weekly/monthly schedules, as our schema doesn't support it
* Added all timezone options that are supported in `Alerting` plugin
* Changing timezone no longer resets cron expression
* Separated timezone component into separate file 


kibana-reports/public/components/report_definitions/report_trigger/timezone.tsx
```diff
@@ -0,0 +1,75 @@
...
+  };
+
+  useEffect(() => {
+    let unmounted = false;
+    if (edit) {
+      httpClientProps
+        .get(`../api/reporting/reportDefinitions/${editDefinitionId}`)
+        .then(async (response) => {
+          if (
+            !unmounted &&
```
joshuali925, line 47: question, why do we need this check in useEffect?

zhongnansu (approved):

akbhatta (approved):

joshuali925 (approved):
LGTM

kibana-reports/public/components/report_definitions/report_trigger/timezone.tsx
davidcui-amzn: the unmounted check prevents potential memory leaks when using `setState`

PR #153, opened 2020-10-27
remove "refresh interval" option from time selector

zhongnansu:
*Issue #, if available:*

*Description of changes:*
remove the "refresh interval" options from time selector in "create report definition", because we don't need it.
image


davidcui-amzn (approved):
LGTM

abbashus (approved):

PR #172, opened 2020-11-05
[Bug Fix] Context menu doesn't show up when switching between tabs from sidebar

zhongnansu:
*Issue #, if available:*
context menu doesn't show up when switching between dashboards & visualizations from sidebar (i.e URL change but no hash change)

*Description of changes:*

When navigating to tabs from Kibana sidebar, it uses `history.pushState`, which doesn't trigger `window.onHashchange`.
Taking solution provided from the following link, to monkey patch event and add event listener. https://stackoverflow.com/questions/4570093/how-to-get-notified-about-changes-of-the-history-via-history-pushstate/4585031


davidcui-amzn (approved):
LGTM- thanks for the fix!

joshuali925 (approved):

akbhatta (approved):

PR #193, opened 2020-11-13
remove create report logic for background job

zhongnansu:
*Issue #, if available:*
Since we don't have user impersonation now, and we only send a link instead of actual file by email, I remove the logic which causing error when security is enabled

*Description of changes:*
- remove create report logic for background job
- leave only delivery report logic
- refine few logging and error handling


kibana-reports/server/executor/createScheduledReport.ts
```diff
@@ -33,33 +34,36 @@ export const createScheduledReport = async (
...
-    } else {
-      // report source can only be one of [saved search, visualization, dashboard]
-      createReportResult = await createVisualReport(
-        reportParams,
-        completeQueryUrl,
-        logger
-      );
-    }
+    // TODO: generate report logic will be added back once we have the user impersonation
+    // if (reportSource === REPORT_TYPE.savedSearch) {
```
davidcui-amzn, line 51: Do we need to leave this commented-out code in?

davidcui-amzn (approved):
LGTM- just one minor comment

akbhatta (approved):

kibana-reports/server/executor/createScheduledReport.ts
zhongnansu: Yes I leave it on purpose, once we have user impersonation added, this code snippet will be reused.

PR #195, opened 2020-11-13
Updated Time Range Parsing in Report Details 

davidcui-amzn:
*Issue #, if available:*
N/A 
*Description of changes:*
Time range in Report Details for `Visualizations` and `Saved searches` from in-context menu would show as undefined. Updated the parsing so it displays correctly. 


kibana-reports/public/components/main/report_details/report_details.tsx
```diff
@@ -121,7 +121,7 @@ export function ReportDetails(props) {
   const parseTimePeriod = (queryUrl: string) => {
     let timeString = queryUrl.substring(
       queryUrl.lastIndexOf('time:'),
-      queryUrl.lastIndexOf('))')
+      queryUrl.indexOf('))')
```
zhongnansu, line 124: This change doesn't seem to be specific for visualization and saved search, but why it was fine with dashboard before this fix?

davidcui-amzn: The url construction for visualization vs saved search vs dashboard is not the same

zhongnansu: I get that it's different but the time range part of the url should be the same, the only difference is dashboard#/view vs discover#/edit, as far as I can remember.

zhongnansu: This solution of locating a substring with `))` seems kind of hack for me. I'd rather use a more generic solution. Could you try this regex expression to match the time field `/time:\(from:.+,to:.+?\)/`

kibana-reports/public/components/main/report_details/report_details.tsx
```diff
@@ -119,19 +119,18 @@ export function ReportDetails(props) {
...
 
   const parseTimePeriod = (queryUrl: string) => {
-    let timeString = queryUrl.substring(
-      queryUrl.lastIndexOf('time:'),
-      queryUrl.indexOf('))')
-    );
+    let timeStringRegex = /time:\(from:.*,to:.*\)/;
+
+    let timeStringRegexArray = timeStringRegex.exec(queryUrl);
+    let timeString = timeStringRegexArray[0];
```
joshuali925, line 125: just a note, you can probably do `const [timeString, fromDateString, toDateString] = queryUrl.match(/time:\(from:(.+),to:(.+?)\)/)` might be easier

also I think the lazy match in `to:.*?\)` is necessary? wouldn't `to:.*\)` match to the last `)`?

joshuali925 (approved):
left question regarding regex

kibana-reports/public/components/main/report_details/report_details.tsx
zhongnansu: image

zhongnansu (approved):
Address the comment before merge

kibana-reports/public/components/main/report_details/report_details.tsx
davidcui-amzn: Good catch - will make change

Issue #200, opened 2020-11-17
Upgrade to ES/Kibana 7.10


zhongnansu:
- [x] Kibana reports plugin
- [x] reports scheduler plugin

PR #205, opened 2020-11-18
[Kibana-reports]Add support for Kibana 7.10.0

zhongnansu:
*Issue #, if available:*
#200 
*Description of changes:*

#### Major change
- remove `target/*` since 7.10 `kbn-plugin-helper` can handle that
- update the context menu button to align with new UI
- move `babel.config.js` into `.babelrc`, since it causes plugin runtime error
- update scripts and dependencies in `package.json` and `kibana-helper.json`

#### TODO: 
- artifact naming format TBD
- github CD workflow need to add valid S3 bucket


kibana-reports/public/components/context_menu/context_menu.js
```diff
@@ -1,3 +1,22 @@
+/*
```
joshuali925, line 1: this is the elasticsearch license, looks like it's accidentally added to some files, ESLint problem?

kibana-reports/server/utils/constants.ts
```diff
@@ -13,4 +13,4 @@
  * permissions and limitations under the License.
  */
 
-export const POLL_INTERVAL = 1000 * 60 * 5; // in milliseconds (5 min)
+export const POLL_INTERVAL = 1000 * 60 * 0.5; // in milliseconds (5 min)
```
joshuali925, line 16: revert to 5 min?

joshuali925 (approved):
LGTM, 2 comments to address

kibana-reports/server/utils/constants.ts
zhongnansu: will do

kibana-reports/public/components/context_menu/context_menu.js
zhongnansu: That's interesting, no idea how it gets added.

kibana-reports/package.json
```diff
@@ -1,49 +1,40 @@
...
     "kbn": "node ../../scripts/kbn",
     "es": "node ../../scripts/es",
     "lint": "eslint .",
-    "start": "plugin-helpers start",
-    "test:server": "plugin-helpers test:server",
-    "test:browser": "plugin-helpers test:browser",
-    "build": "plugin-helpers build",
-    "test:jest": "../../node_modules/.bin/jest --config ./test/jest.config.js",
-    "test": "jest --config ./test/jest.config.js",
+    "start": "node ../../scripts/plugin_helpers start",
```
yizheliu-amazon, line 15: You may add `plugin-helpers` script to avoid multiple  `node ../../scripts/plugin_helpers` below
```
"plugin-helpers": "node ../../scripts/plugin_helpers",

Comment (zhongnansu) [reply kibana-reports/package.json:15]: good idea, thx!

**Comment by yizheliu-amazon** [comment]
have you verified the kibana-report artifact can be installed to binary kibana, and it is working as expected when kibana is started?

File: kibana-reports/package.json (hunk 3 above)
Comment (zhongnansu) [reply kibana-reports/package.json:15]: I look into some details, I can add the plugin_helper as script, but to avoid multiple `node ../../scripts/plugin_helpers`, what I need to do is actually variable reference in `package.json`. There seems like only one way to do it.
Ref: https://brianchildress.co/variables-in-package-json/

```
"config": {
    "plugin_helpers": "node ../../scripts/plugin_helpers"
  },
  "scripts": {
    ...
    "start": "$npm_package_config_plugin_helpers start",
    ...
}

```

Comment (zhongnansu) [reply kibana-reports/package.json:15]: It looks to me that it's still very verbose, and the console output is less readable. You won't see the actual script getting executed, instead you see
image

**Comment by zhongnansu** [comment]
> have you verified the kibana-report artifact can be installed to binary kibana, and it is working as expected when kibana is started?

@yizheliu-amazon Yes, it works as expected for reporting plugin

File: kibana-reports/package.json (hunk 3 above)
Comment (joshuali925) [reply kibana-reports/package.json:15]: https://github.com/joshuali925/trace-analytics/blob/7.10/package.json#L12-L17
you can try this, $env is not cross platform

Comment (zhongnansu) [reply kibana-reports/package.json:15]: this is helpful, thx : )

Comment (ftianli-amzn) [reply kibana-reports/package.json:15]: I‘m not sure if it works for you, I think what Yizhe said is like here https://github.com/elastic/kibana/tree/master/packages/kbn-plugin-helpers#installation

File: kibana-reports/package.json [hunk 4]
```diff
@@ -1,49 +1,37 @@
...
-    "test:server": "plugin-helpers test:server",
-    "test:browser": "plugin-helpers test:browser",
-    "build": "plugin-helpers build",
-    "test:jest": "../../node_modules/.bin/jest --config ./test/jest.config.js",
-    "test": "jest --config ./test/jest.config.js",
-    "cypress:run": "./node_modules/.bin/cypress run",
-    "cypress:open": "./node_modules/.bin/cypress open"
+    "start": "yarn plugin_helpers start",
+    "test:server": "yarn plugin_helpers test:server",
+    "test:browser": "yarn plugin_helpers test:browser",
```
yizheliu-amazon, line 17: I guess these `start/test:server/test:browser`are not supported by plugin-helper anymore. They are supported in only [old version](https://github.com/elastic/kibana/tree/v7.4.1/packages/kbn-plugin-helpers) now. You may remove them if not used.

ftianli-amzn: They are not supported, but I'm looking forward to get a alternative way for the original function.

zhongnansu: good catch, will do

akbhatta (approved):

PR #212, opened 2020-11-23
Disable api calls that uses Kibana default internal user

zhongnansu:
*Issue #, if available:*
We have security limitations for calling ES api with kibana default user in `kibana.yml`, which is the kibana server role. This PR removes the affected code and propose a solution.

*Description of changes:*
- disable poller, since in the backend, it has already mark every scheduled task as scueess
- add a new Kibana API `POST .../reporting/genreateReport/{reportDefinitionId}`, which calls the ES API 
```
POST .../_reports/on_demand/{reportDefinitionId}
body: {reportDefinitionId: reportDefinitionId}
```
- disable all api calls to update report status
- change `.../generateReport/{reportId}` from a `POST` request to a `GET` request, because this call will not generate resource to the server, it just generate reports on the fly without saving anything. But `.../generateReport/{reportId}` does create a new instance to the `report` index.
- TODO: UI now needs to call the above new API for the following instead of old `POST ..../generateReport` with a request body of complete parameters.
  - Generate button from report definition detail page, and the icon download as well
  - the second API call for the "create on-demand report definition" workflow

### DIFF
--- a/kibana-reports/public/components/main/main_utils.tsx
+++ b/kibana-reports/public/components/main/main_utils.tsx
@@ -179,7 +179,7 @@ export const generateReportById = async (
   handlePermissionsMissingToast
 ) => {
   await httpClient
-    .post(`../api/reporting/generateReport/${reportId}`)
+    .get(`../api/reporting/generateReport/${reportId}`)
     .then(async (response) => {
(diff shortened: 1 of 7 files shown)


davidcui-amzn (approved):
LGTM

akbhatta (approved):

PR #213, opened 2020-11-24
Use new API to Generate Reports from Existing Definitions

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
Refactored to call new API for generating reports from existing definitions- instead of passing in the entire metadata of creating a report, we only have to pass in the `reportDefinitionId` now. 

Usages:
* Creating a report after an on-demand report definition is created
* The `Generate report` button on the `Report definition details` page for an on-demand definition.
* The download icon under `File format` within `Report definition details`. 

### DIFF
--- a/kibana-reports/public/components/main/report_definition_details/report_definition_details.tsx
+++ b/kibana-reports/public/components/main/report_definition_details/report_definition_details.tsx
@@ -463,10 +463,6 @@ export function ReportDefinitionDetails(props) {
   };
 
   const generateReportFromDetails = async () => {
-    let duration =
-      reportDefinitionRawResponse.report_definition.report_params.core_params
-        .time_duration;
-    const fromDate = getRelativeStartDate(duration);
     const { httpClient } = props;
     let generateReportSuccess = await generateReportFromDefinitionId(
       reportDefinitionId,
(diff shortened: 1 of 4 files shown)


akbhatta (approved):

kibana-reports/public/components/main/report_definition_details/report_definition_details.tsx
```diff
@@ -467,17 +467,9 @@ export function ReportDefinitionDetails(props) {
       reportDefinitionRawResponse.report_definition.report_params.core_params
         .time_duration;
     const fromDate = getRelativeStartDate(duration);
```
zhongnansu, line 469: Do we still need all these?

davidcui-amzn: No, good catch. Will remove

zhongnansu: Can you verify if it's the same situation for logic in create_report_defintion.ts? I always have an impression that lots of code can be removed with this new API

davidcui-amzn: Do you mean `create_report_definition.tsx`? If so, yes. From the code change in the PR, the only parameter that is created is a `reportDefinitionId` constant based on the response from the `Create` API call

zhongnansu (approved):
LGTM, merge after github action test and build is passed

PR #221, opened 2020-11-25
Added Loading Modal to Details pages

davidcui-amzn:
*Issue #, if available:*
N/A
*Description of changes:*
Extract `GenerateReportLoadingModal` and added it to `Report details` and `Report definition details` for file format download. 

### DIFF
--- a/kibana-reports/public/components/main/loading_modal/index.ts
+++ b/kibana-reports/public/components/main/loading_modal/index.ts
@@ -1,0 +1,16 @@
+/*
+ * Copyright 2020 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License").
+ * You may not use this file except in compliance with the License.
+ * A copy of the License is located at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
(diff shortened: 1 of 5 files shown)


akbhatta (approved):

joshuali925 (approved):

kibana-reports/public/components/main/loading_modal/index.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * or in the "license" file accompanying this file. This file is distributed
+ * on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
+ * express or implied. See the License for the specific language governing
+ * permissions and limitations under the License.
+ */
+
+ export { GenerateReportLoadingModal } from './loading_modal';
```
zhongnansu, line 16: why need this, you already export this from `loading_modal.tsx`

davidcui-amzn: So when we export it in other files we don't need to write `import from '../../loading_modal/loading_modal`, eliminates need for nested import

zhongnansu: I think loading_modal is like a util components, which may better be placed under `utils` for other components to call. Otherwise, if you write a index.ts for every small component, the project will have too many files. Just my 2 cents, may refer to some project structure standard later when we do the refactor

PR #222, opened 2020-11-25
Use puppeteer-core with custom chromium instead of puppeteer

joshuali925:
*Issue #, if available:*

*Description of changes:*
- remove puppeteer dependency, use puppeteer-core with custom chromium binary instead
- change plugin id from `opendistroKibanaReports` to `opendistroReportsKibana` for consistency
- add chromium path parameter to `createVisualReport` for jest testing
- update github workflows to build artifacts for windows and linux
- add s3 bucket path in release workflow
- use `path.posix.normalize` instead of `path.normalize` when verifying URL to accommodate Windows

## Caveats
- Chromium binary has to be downloaded separately if running from source, but will be packed by github actions in the zip artifact
- <del>Chromium path defaults to `./plugins/opendistroReportsKibana/.chromium/headless_shell` which works for the zip artifact, but if running from source it should be `./plugins/kibana-reports/.chromium/headless_shell`</del> (Fixed by #236 )
- Ubuntu needs additional dependencies to run chromium 
```
sudo apt install -y libnss3-dev fonts-liberation libfontconfig1
```
- RedHat/CentOS needs additional dependencies to run chromium
```
sudo yum install -y libnss3.so xorg-x11-fonts-100dpi xorg-x11-fonts-75dpi xorg-x11-utils xorg-x11-fonts-cyrillic xorg-x11-fonts-Type1 xorg-x11-fonts-misc fontconfig freetype
```

### DIFF
--- a/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
+++ b/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
@@ -22,6 +22,7 @@ import {
   REPORT_TYPE,
   FORMAT,
   SELECTOR,
+  CHROMIUM_PATH,
 } from '../constants';
 import { getFileName } from '../helpers';
 import { CreateReportResultType } from '../types';
@@ -33,7 +34,8 @@ export const createVisualReport = async (
   reportParams: ReportParamsSchemaType,
   queryUrl: string,
   logger: Logger,
-  cookie?: SetCookie
+  cookie?: SetCookie,
+  chromiumPath = CHROMIUM_PATH
 ): Promise<CreateReportResultType> => {
   const {
     core_params,
@@ -66,7 +68,7 @@ export const createVisualReport = async (
      * https://github.com/puppeteer/puppeteer/blob/main/docs/troubleshooting.md#setting-up-chrome-linux-sandbox
      */
     args: ['--no-sandbox', '--disable-setuid-sandbox'],
-    executablePath: './plugins/kibana-reports/.chromium/headless_shell',
+    executablePath: chromiumPath,
   });
   const page = await browser.newPage();
   page.setDefaultNavigationTimeout(0);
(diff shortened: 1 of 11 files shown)


kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
```diff
@@ -33,7 +34,8 @@ export const createVisualReport = async (
   reportParams: ReportParamsSchemaType,
   queryUrl: string,
   logger: Logger,
-  cookie?: SetCookie
+  cookie?: SetCookie,
+  chromiumPath = CHROMIUM_PATH
```
zhongnansu, line 38: can you put the optional parameter as the last parameter? So that in the test file, I don't think you need to pass in `undefined` anymore

zhongnansu:
Should we provide the different chromium version somewhere for developers? mac, windows and linux?

kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
joshuali925: I don't know where the `cookie` parameter is used, if I swap them whenever you pass in cookie you'll have to pass `undefined` or chromium path before cookie

joshuali925:
> Should we provide the different chromium version somewhere for developers? mac, windows and linux?

they are currently in a release under reporting https://github.com/opendistro-for-elasticsearch/kibana-reports/releases/tag/chromium-1.12.0.0

kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
zhongnansu: In your code, the chromium parameter already has an initializer which uses `CHROMIUM_PATH`, besides test, I don't think we need to pass in different value for the path

zhongnansu: Oh I see your point, that won't work. Never mind, ignore my comment

joshuali925: BTW I tried to mock `CHROMIUM_PATH` and used `jest.requireActual` to import other constants in the test file. The constants are correct in the test file, but in `visualReportHelper.ts` only `CHROMIUM_PATH` is correct. 

That's why I used the optional parameter workaround, if you are more familiar with jest you can try to mock the constant which would be a better solution

zhongnansu (approved):

davidcui1225 (approved):
LGTM

PR #224, opened 2020-11-25
[Bug fix] Use default max size to call getAll ES API

zhongnansu:
*Issue #, if available:*

The default query size in ES side is set to 100, should pass in parameter if we want to retrieve more data

*Description of changes:*
- From Kibana side, add `fromIndex` and `maxItems` as optional params, and use default value `10000` for `maxItems`.
- Test with 10000 report instances added. Browser response time is within 4 secs. It's tolerable for the initial release.
- TODO: need to implement pagination in the EUI, currently the table eui component is [EuiInMemoryTable](https://elastic.github.io/eui/#/tabular-content/in-memory-tables), which requires all possible data to be loaded. Need to consider use low level components such as `EuiBasicTable` along with `EuiPagination`, `EuiSearchBar`

#### Others
- set enabled field to be true when switching from on-demand to scheduled report definition

### DIFF
--- a/kibana-reports/public/components/main/main.tsx
+++ b/kibana-reports/public/components/main/main.tsx
@@ -36,7 +36,6 @@ import {
   permissionsMissingToast,
   permissionsMissingActions,
 } from '../utils/utils';
-import { HttpSetup } from '../../../../../src/core/public/http/types';
 
 const reportCountStyles: CSS.Properties = {
   color: 'gray',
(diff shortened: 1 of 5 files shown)


akbhatta (approved):

joshuali925 (approved):

davidcui-amzn (approved):
LGTM

PR #228, opened 2020-11-30
Update workflow to build artifact for ARM64

joshuali925:
*Issue #, if available:*

*Description of changes:*
- adds linux arm64 builds in addition to linux x64 and windows in #222 
- renamed chromium binaries to include architecture information
- add `puppeteer-core` as it's not included in kibana-oss

Artifact names:
`opendistroReportsKibana-1.12.0.0-linux-x64.zip`
`opendistroReportsKibana-1.12.0.0-linux-arm64.zip`
`opendistroReportsKibana-1.12.0.0-windows-x64.zip`

### DIFF
--- a/.github/workflows/kibana-reports-release-workflow.yml
+++ b/.github/workflows/kibana-reports-release-workflow.yml
@@ -37,50 +37,58 @@ jobs:
           node-version: "10.22.1"
 
       - name: Move Kibana Reports to Plugins Dir
-        run: mv kibana-reports kibana/plugins/opendistroReportsKibana
+        run: mv kibana-reports kibana/plugins/${{ env.PLUGIN_NAME }}
+
+      - name: Add Chromium Binary to Reporting for Testing
+        run: |
+          sudo apt install -y libnss3-dev fonts-liberation libfontconfig1
+          cd kibana/plugins/${{ env.PLUGIN_NAME }}
+          wget https://github.com/opendistro-for-elasticsearch/kibana-reports/releases/download/chromium-1.12.0.0/chromium-linux-x64.zip
+          unzip chromium-linux-x64.zip
+          rm chromium-linux-x64.zip
 
       - name: Kibana Plugin Bootstrap
         uses: nick-invision/retry@v1
         with:
           timeout_minutes: 30
           max_attempts: 3
-          command: cd kibana/plugins/opendistroReportsKibana; yarn kbn bootstrap
+          command: cd kibana/plugins/${{ env.PLUGIN_NAME }}; yarn kbn bootstrap
 
       - name: Test
         uses: nick-invision/retry@v1
         with:
           timeout_minutes: 30
           max_attempts: 3
-          command: cd kibana/plugins/opendistroReportsKibana; yarn test
+          command: cd kibana/plugins/${{ env.PLUGIN_NAME }}; yarn test
 
       - name: Build Artifact
         run: |
-          cd kibana/plugins/opendistroReportsKibana
+          cd kibana/plugins/${{ env.PLUGIN_NAME }}
           yarn build
 
           cd build
-          mkdir -p ./{linux-x64,linux-arm64,windows-x64}/kibana/opendistroReportsKibana
+          mkdir -p ./{linux-x64,linux-arm64,windows-x64}/kibana/${{ env.PLUGIN_NAME }}
           cp ./${{ env.PLUGIN_NAME }}-*.zip ./linux-x64/${{ env.PLUGIN_NAME }}-${{ env.OD_VERSION }}-linux-x64.zip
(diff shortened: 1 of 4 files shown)


davidcui-amzn (approved):
LGTM

.github/workflows/kibana-reports-release-workflow.yml
```diff
@@ -59,20 +59,28 @@ jobs:
           yarn build
 
           cd build
-          mkdir -p ./{linux,windows}/kibana/opendistroReportsKibana
-          cp ./${{ env.PLUGIN_NAME }}-*.zip ./linux/${{ env.PLUGIN_NAME }}-${{ env.OD_VERSION }}-linux-x64.zip
-          mv ./${{ env.PLUGIN_NAME }}-*.zip ./windows/${{ env.PLUGIN_NAME }}-${{ env.OD_VERSION }}-windows-x64.zip
+          mkdir -p ./{linux-x64,linux-arm64,windows-x64}/kibana/opendistroReportsKibana
```
zhongnansu, line 62: use env var for the plugin name?

.github/workflows/kibana-reports-test-and-build-workflow.yml
```diff
@@ -56,32 +56,46 @@ jobs:
           yarn build
 
           cd build
-          mkdir -p ./{linux,windows}/kibana/opendistroReportsKibana
-          cp ./${{ env.PLUGIN_NAME }}-*.zip ./linux/${{ env.PLUGIN_NAME }}-${{ env.OD_VERSION }}-linux-x64.zip
-          mv ./${{ env.PLUGIN_NAME }}-*.zip ./windows/${{ env.PLUGIN_NAME }}-${{ env.OD_VERSION }}-windows-x64.zip
+          mkdir -p ./{linux-x64,linux-arm64,windows-x64}/kibana/opendistroReportsKibana
```
zhongnansu, line 59: same here

kibana-reports/yarn.lock
```diff
@@ -863,6 +863,13 @@ acorn@^6.0.1, acorn@^6.0.4, acorn@^6.4.1:
   resolved "https://registry.yarnpkg.com/acorn/-/acorn-6.4.2.tgz#35866fd710528e92de10cf06016498e47e39e1e6"
   integrity sha512-XtGIhXwF8YM8bJhGxG5kXgjkEuNGLTkoYqVE+KMR+aspr4KGYmKYg7yUe3KghyQ9yheNwLnjmzh/7+gfDBmHCQ==
 
+agent-base@^4.3.0:
+  version "4.3.0"
+  resolved "https://registry.yarnpkg.com/agent-base/-/agent-base-4.3.0.tgz#8165f01c436009bccad0b1d122f05ed770efc6ee"
+  integrity sha512-salcGninV0nPrwpGNn4VTXBb1SOuXQBiqbrNXoeizJsHrsL6ERFM2Ne3JUSBWRE6aeNJI2ROP/WEEIDUiDe3cg==
```
zhongnansu, line 869: Are those new dependencies in `yarn.lock` introduced by puppeteer-core ?

.github/workflows/kibana-reports-test-and-build-workflow.yml
```diff
@@ -32,9 +32,9 @@ jobs:
         run: |
```
zhongnansu, line 32: This step is missing in the release flow

zhongnansu:
Do we have plan to add chromium artifacts for macOS?

.github/workflows/kibana-reports-test-and-build-workflow.yml
joshuali925: Added, although I think we only need to test in CI, shouldn't need to test in CD?

kibana-reports/yarn.lock
joshuali925: yes

.github/workflows/kibana-reports-test-and-build-workflow.yml
joshuali925: updated, thanks

.github/workflows/kibana-reports-test-and-build-workflow.yml
zhongnansu: it may be a requirement from infra team, guess all the plugins are doing this way

zhongnansu (approved):

joshuali925:
> Do we have plan to add chromium artifacts for macOS?

I don't think so, ODFE doesn't have a mac version

zhongnansu:
> > Do we have plan to add chromium artifacts for macOS?
> 
> I don't think so, ODFE doesn't have a mac version

developers may want it

PR #239, opened 2020-12-03
Add Custom Common Time Ranges

davidcui1225:
*Issue #, if available:*
N/A
*Description of changes:*
The default custom time ranges in the date picker end up with invalid time ranges. Replacing them with custom time ranges that don't violate our time range:
* Today so far
* Week to date
* Month to date
* Year to date 

Screen Shot 2020-12-03 at 9 50 33 AM


joshuali925 (approved):
Why do we limit time range to be <= now? And what happens if users manually select a future date?

davidcui1225:
> Why do we limit time range to be <= now? And what happens if users manually select a future date?

There is no data to be displayed if the time range ends in the future, so we do not support it. If a future date is selected, the Time Range error toast will display, as well as the client-side input validation not allowing the definition to be created.

zhongnansu (approved):

PR #240, opened 2020-12-03
Fix the time range display issue(timezone) on visual report

zhongnansu:
*Issue #, if available:*

We recently decided to add the time range to the visual report(PDF/PNG), that introduced one issue. The time range always displays as UTC time. Because puppeteer launches the chromium from server side, so by default it’s using UTC. Ideally it should be user’s local timezone

*Description of changes:*
- Add timezone as a query parameter for all APIs that returns actual report binary. Grab the timezone from browser (client side)
- add env var to puppeteer launch browser process, which is visible to Chromium. So we can simulate a user accessing a dashboard from his timezone and rely on the default Kibana UI setting to use the browser timezone for time range display.

#### Notes
I also tried other solutions
1. provide time string with offset in the `query_url`, but Kibana fails to recognize that. It seems it only accepts zulu time.
2. Add an extra timezone field. We are not going with this solution due to the concern of backward compatibility. The accepted solution doesn't introduce any change to the schema

### DIFF
--- a/kibana-reports/server/routes/utils/converters/backendToUi.ts
+++ b/kibana-reports/server/routes/utils/converters/backendToUi.ts
@@ -50,7 +50,8 @@ import {
 import moment from 'moment';
 
 export const backendToUiReport = (
-  backendReportInstance: BackendReportInstanceType
+  backendReportInstance: BackendReportInstanceType,
+  timezone?: string
 ): ReportSchemaType => {
   const {
     inContextDownloadUrlPath,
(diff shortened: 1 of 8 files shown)


akbhatta (approved):

kibana-reports/server/routes/utils/converters/backendToUi.ts
```diff
@@ -203,6 +203,7 @@ const getUiQueryUrl = (
 ) => {
   const timeFrom = moment(beginTimeMs).toISOString();
   const timeTo = moment(endTimeMs).toISOString();
+
```
davidcui1225: Was this meant to be included in the PR?

davidcui1225 (approved):
LGTM

joshuali925 (approved):

kibana-reports/server/routes/utils/converters/backendToUi.ts
zhongnansu: fixed

PR #241, opened 2020-12-03
Bug Fixes in UI

davidcui1225:
*Issue #, if available:*
N/A
*Description of changes:*
Addressed low severity bug fixes.
* `Report details` and `Report definition details` display a dash line for header and footer like all other blank fields instead of showing nothing
* In-context menu text was changed to be more extensible for all report sources
* Cron help link in `Report trigger` opens in a new tab

### DIFF
--- a/kibana-reports/public/components/main/report_definition_details/report_definition_details.tsx
+++ b/kibana-reports/public/components/main/report_definition_details/report_definition_details.tsx
@@ -336,10 +336,10 @@ export function ReportDefinitionDetails(props) {
       // TODO: need better display
       timePeriod: moment.duration(timeDuration).humanize(),
       fileFormat: reportFormat,
-      reportHeader: reportParams.core_params.hasOwnProperty('header')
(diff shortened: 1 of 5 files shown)


akbhatta (approved):

kibana-reports/public/components/main/report_definition_details/report_definition_details.tsx
```diff
@@ -336,10 +336,10 @@ export function ReportDefinitionDetails(props) {
       // TODO: need better display
       timePeriod: moment.duration(timeDuration).humanize(),
       fileFormat: reportFormat,
-      reportHeader: reportParams.core_params.hasOwnProperty('header')
+      reportHeader: reportParams.core_params.hasOwnProperty('header') && reportParams.core_params.header != ""
```
joshuali925, line 339: just a note it's better to use `!==` in general so no implicit conversion is done, although in this case it doesn't seem to matter

joshuali925 (approved):

PR #242, opened 2020-12-04
add double dots to relative url in fetch()

zhongnansu:
*Issue #, if available:*
in-context menu download fails when 'server.basepath' in kibana.yml is configured, by default it's empty string.

*Description of changes:*
- Add 2 dots to use relative url, just like how we call API in other places such as create report definition, getAll reports/report definition


joshuali925 (approved):
thanks for the fix

davidcui1225 (approved):
LGTM

PR #244, opened 2020-12-10
Optimize selectors for DOM operation to reduce possible version compatibility issue 

zhongnansu:
*Issue #, if available:*

The current DOM element selector we use is not generic enough, so it may lead to more manual work during each version compatibility check. Some html element of Kibana dashboard with minor update will easily break the report creation logic.

*Description of changes:*
Instead of using a specific class/id selector of a specific element in a specific kibana version, I tried to come up with a generic selector while we operate the DOM.


joshuali925 (approved):

davidcui1225 (approved):
LGTM

PR #245, opened 2020-12-11
Backport from branch opendistro-1.12.0.0 to 7.9.1

zhongnansu:
*Issue #, if available:*

*Description of changes:*
- Cherry-pick commits except #203 #205 which is a version upgrade from 7.9.1 to 7.10.0
- fix some conflicts that relates to 7.10 in other commits
- change the constant `LOCAL_HOST` from `5601` to `9200` for AES
- add a constant `BASE_PATH = "/_plugin/kibana"` for AES
- use patch release `common-utils:1.11.1.0` in `build.gradle`

#### TODO
- come up a solution to close the code gap between AES and OD. Need to use plugin-config for at least `port` and `basePath`. This will also remove the limitation of current reporting on OD, where reporting only works if user use default `server.basePath` and `server.port` in Kibana.yml. The code change should be pushed to `dev` branch and backport to `7.9.1` branch if needed.


davidcui1225 (approved):
LGTM

joshuali925 (approved):

joshuali925:
@zhongnansu looks like basepath can be retrieved from `core.http.basePath.get(request)` or `core.http.basePath.serverBasePath` 
https://github.com/elastic/kibana/blob/v7.9.1/src%2Fcore%2Fserver%2Fhttp%2Fbase_path_service.ts#L31-L36

and port is `core.http.getServerInfo().port`
https://github.com/elastic/kibana/blob/v7.9.1/src%2Fcore%2Fserver%2Fhttp%2Ftypes.ts#L272-L275

if they don't work maybe read from frontend url or ping a few possible `localhost:<port>/<path>/api/status` to see which one responds?

zhongnansu:
> @zhongnansu looks like basepath can be retrieved from `core.http.basePath.get(request)` or `core.http.basePath.serverBasePath`
> https://github.com/elastic/kibana/blob/v7.9.1/src%2Fcore%2Fserver%2Fhttp%2Fbase_path_service.ts#L31-L36
> 
> and port is `core.http.getServerInfo().port`
> https://github.com/elastic/kibana/blob/v7.9.1/src%2Fcore%2Fserver%2Fhttp%2Ftypes.ts#L272-L275
> 
> if they don't work maybe read from frontend url or ping a few possible `localhost:<port>/<path>/api/status` to see which one responds?



> @zhongnansu looks like basepath can be retrieved from `core.http.basePath.get(request)` or `core.http.basePath.serverBasePath`
> https://github.com/elastic/kibana/blob/v7.9.1/src%2Fcore%2Fserver%2Fhttp%2Fbase_path_service.ts#L31-L36
> 
> and port is `core.http.getServerInfo().port`
> https://github.com/elastic/kibana/blob/v7.9.1/src%2Fcore%2Fserver%2Fhttp%2Ftypes.ts#L272-L275
> 
> if they don't work maybe read from frontend url or ping a few possible `localhost:<port>/<path>/api/status` to see which one responds?

yes, basePath can be retrieved that way, only with request context, so the scheduled jobs won't work, because we are not saving it into database. Besides we also need `port`, which can't be retrieved.

zhongnansu:
> core.http.basePath.serverBasePath

Good point! I wasn't able to find how to access port before. Your solution sounds promising, the only thing left is to rebuild the input validation for the relative url. Do you know if there is a way to expose basePath to UI, or access basePath from UI code? Basically, we need to clean the url from UI before passing to server, so that the old input validation will work and robust with minor modification

joshuali925:
That will be part of the `basename` [here](https://github.com/opendistro-for-elasticsearch/kibana-reports/blob/7.9.1/kibana-reports/public/components/app.tsx#L63). I think it would be something like `/basepath/app/opendistro-reports`, you can get Kibana server.basepath if you remove the plugin path.

ylasri:
Hi @zhongnansu 

Only some parameters from Kibana.yml are exposed via CoreSetup

- server.port
- server.host
- server.name
- server.basePath

Theses parameters can be accessed as follow from any server method that has CoreSetup as a parameter

```
  const { protocol, port, hostname, name } = core.http.getServerInfo();
  const basePath = core.http.basePath.serverBasePath;
```

```
core.http.getServerInfo() is type of HttpServerInfo 

/** @public */
export interface HttpServerInfo {
  /** The name of the Kibana server */
  name: string;
  /** The hostname of the server */
  hostname: string;
  /** The port the server is listening on */
  port: number;
  /** The protocol used by the server */
  protocol: 'http' | 'https' | 'socket';
}
```

The CoreSetup can instance is the one used at the server level plugin initilization

```
public setup(core: CoreSetup) {
  ....
  //Register server side APIs
  registerRoutes(router, core);
  ....
}
```

PR #261, opened 2020-12-17
Reports Table Backport Changes

davidcui1225:
*Issue #, if available:*
N/A
*Description of changes:*
* In-context menu reports now are named after the title of the report source instead of just `On_demand report`
* Removed report definitions table
* Removed reports table filters
* Added report ID field with report name + uuid
* Removed option to go to `Report details` page
* Changed source field to use report name vs source type
* Add time period field 

Screen Shot 2020-12-16 at 5 37 06 PM


kibana-reports/package.json
```diff
@@ -32,6 +32,7 @@
     "@types/dompurify": "^2.0.4",
     "@types/jsdom": "^16.2.4",
     "@types/react-addons-test-utils": "^0.14.25",
+    "@types/uuid": "^8.3.0",
```
zhongnansu: remove all new uuid related dependencies, you don't need those for `import uuidv1() from uuid/v1`

kibana-reports/public/components/main/reports_table.tsx
```diff
@@ -37,6 +38,8 @@ import {
   humanReadableDate,
   generateReportById,
 } from './main_utils';
+import dateMath from '@elastic/datemath';
+import uuidv4 from 'uuid/v4';
```
zhongnansu, line 41: in the server side we use v1

zhongnansu (changes requested):
please address the comments

kibana-reports/public/components/main/reports_table.tsx
davidcui1225: v1 is based on timestamp, v4 is random. I don't think it will make a big difference

kibana-reports/package.json
davidcui1225: will do

kibana-reports/public/components/main/reports_table.tsx
```diff
@@ -14,6 +14,7 @@
  */
 
 import React, { Fragment, useState } from 'react';
+import ReactDOM from "react-dom";
```
joshuali925: why is reactdom needed?

davidcui1225: It's not, will remove- added it as a try while debugging

joshuali925 (approved):

zhongnansu (approved):

zhongnansu:
Just found an issue, you are genrating the uuid on the fly, it changes when I refresh the report table. This name should be persistent, and saved as the report name field in the schema. @davidcui-amzn 

Also could you clean the `yarn.lock`, by `yarn kbn clean` and `yarn kbn bootstrap`? Those `uuid` dependencies are still there.

PR #263, opened 2020-12-17
Fix UUID Generation

davidcui1225:
*Issue #, if available:*
N/A
*Description of changes:*
* Make UUID part of report name in schema so it's not generated on-fly in reports table
* Make time period column text same size as rest of columns
* Update snapshot
* Report source field displays report name with UUID parsed out


kibana-reports/public/components/context_menu/context_menu.js
```diff
@@ -93,7 +94,7 @@ const generateInContextReport = (
     time_to: timeRanges.time_to.valueOf(),
     report_definition: {
       report_params: {
-        report_name: $('span.euiBreadcrumb').prop('title'),
+        report_name: $('span.euiBreadcrumb').prop('title') + '_' + uuidv4(),
```
zhongnansu, line 97: Do we have snapshot testing for in-context menu UI? Or is there a way to add?

davidcui1225: We can add snapshot testing- but this line `97` highlighted will not be tested even if we add in-context UI snapshots

zhongnansu (approved):

joshuali925 (approved):

PR #280, opened 2020-12-29
Fixed Edit Report Definition Trigger Type Pre-fill

davidcui1225:
*Issue #, if available:*
N/A
*Description of changes:*
When going to the `Edit report definition` page for a recurring definition, report definitions that had a `Daily` configuration would always be pre-filled with an interval of `Every 1 minute`. Now, daily definitions correctly fill to `Daily` in `Report trigger`. 


zhongnansu (approved):

abbashus (approved):
Thanks for the change

PR #284, opened 2021-01-05
Add semaphore to block on puppeteer chromium execution

joshuali925:
*Issue #, if available:*

*Description of changes:*
- currently number of resources for the semaphore is 1
- semaphore times out and front end fails after 180 seconds, if timeout UI will display this toast:
image
- for on demand it displays this toast depending on report source:
image

### DIFF
--- a/kibana-reports/public/components/main/main_utils.tsx
+++ b/kibana-reports/public/components/main/main_utils.tsx
@@ -198,6 +198,8 @@ export const generateReportById = async (
       console.log('error on generating report by id:', error);
       if (error.body.statusCode === 403) {
         handlePermissionsMissingToast();
+      } else if (error.body.statusCode === 503) {
+        handleErrorToast('Server busy, please try again later.')
       } else {
         handleErrorToast();
       }
--- a/kibana-reports/public/components/main/main_utils.tsx
+++ b/kibana-reports/public/components/main/main_utils.tsx
@@ -199,7 +199,10 @@ export const generateReportById = async (
       if (error.body.statusCode === 403) {
         handlePermissionsMissingToast();
       } else if (error.body.statusCode === 503) {
-        handleErrorToast('Server busy, please try again later.')
+        handleErrorToast(
+          'Error generating report.',
+          `Timed out generating report ID ${reportId}. Try again later.`
+        );
       } else {
         handleErrorToast();
       }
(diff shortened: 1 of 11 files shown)


kibana-reports/server/routes/lib/createReport.ts
```diff
@@ -102,13 +105,18 @@ export const createReport = async (
...
         });
       }
-      createReportResult = await createVisualReport(
-        reportParams,
-        completeQueryUrl,
-        logger,
-        cookieObject,
-        timezone
-      );
+      const [value, release] = await semaphore.acquire();
```
penghuo, line 108: what if the acquire failed/timeout? is there retry avaliable?

penghuo:
Do we have the UT/IT to cover the change?

kibana-reports/server/plugin.ts
```diff
@@ -50,9 +51,11 @@ export class OpendistroKibanaReportsPlugin
       OpendistroKibanaReportsPluginStart
     > {
   private readonly logger: Logger;
+  private readonly semaphore: SemaphoreInterface;
 
   constructor(initializerContext: PluginInitializerContext) {
     this.logger = initializerContext.logger.get();
+    this.semaphore = withTimeout(new Semaphore(1), 30000, new Error('timeout'));
```
akbhatta: I think 30 sec is too less a timeout. may be 180 seconds?

kibana-reports/server/routes/lib/createReport.ts
akbhatta: I think retrying semaphore not required. we can just increase timeout to required value.

joshuali925: If request fails or timeout, client side would get an error. I think increasing timeout is better than client auto retry. I can improve the error message to indicate it's a timeout on UI

kibana-reports/server/plugin.ts
joshuali925: changed to 180 seconds

akbhatta (approved):

zhongnansu (approved):

kibana-reports/public/components/main/main_utils.tsx
```diff
@@ -198,6 +198,11 @@ export const generateReportById = async (
       console.log('error on generating report by id:', error);
       if (error.body.statusCode === 403) {
         handlePermissionsMissingToast();
+      } else if (error.body.statusCode === 503) {
+        handleErrorToast(
+          'Error generating report.',
+          `Timed out generating report ID ${reportId}. Try again later.`
```
penghuo, line 204: does 503 always because of timeout?

PR #292, opened 2021-01-07
Update timeRangeMatcher to avoid create report failure

zhongnansu:
*Issue #, if available:*
- Fail to generate report on certain visualization items

*Description of changes:*
- Root cause: the timeRangeMatcher fails on certain input urls. So the `generate report` API call fails to make from client to server
- Solution: update the regex solves the issue


joshuali925 (approved):
thanks for the fix

anirudha (approved):

PR #293, opened 2021-01-07
Fix the selected fields issue in csv report

zhongnansu:
*Issue #, if available:*

selected fields for saved search not showing as expected on csv reports. issues, all the columns are included instead of the selected fields only.

*Description of changes:*
#### Root cause
There is a global variable `metadata.selectedFields` being used in the csv report logic, which doesn't reset when user switches saved searches.
#### Solution
create scope variable for collecting selected columns and assign to the global variable `metadata`

### DIFF
--- a/kibana-reports/server/routes/utils/dataReportHelpers.ts
+++ b/kibana-reports/server/routes/utils/dataReportHelpers.ts
@@ -37,14 +37,16 @@ export var metaData = {
 
 // Get the selected columns by the user.
 export const getSelectedFields = async (columns) => {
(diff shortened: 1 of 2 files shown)


dai-chen (approved):
Thanks for the fix! Btw, does the UT for this file still work or not?

zhongnansu:
> Thanks for the fix! Btw, does the UT for this file still work or not?

Yes, the github action passed, which means the tests are fine

joshuali925 (approved):
LGTM

PR #295, opened 2021-01-08
Context menu popout & Report definitions Toast fixes

davidcui1225:
*Issue #, if available:*
N/A
*Description of changes:*
* Download in progress modal disappears after any response is received, so it will not remain on-screen after operation is done in error cases
* Commented out report definition toasts since we don't display the report definitions table 

### DIFF
--- a/kibana-reports/public/components/context_menu/context_menu.js
+++ b/kibana-reports/public/components/context_menu/context_menu.js
@@ -121,8 +121,8 @@ const generateInContextReport = (
     }
   )
     .then((response) => {
+      $('#reportGenerationProgressModal').remove();
       if (response.status === 200) {
-        $('#reportGenerationProgressModal').remove();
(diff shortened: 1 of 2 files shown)


zhongnansu:
Do you confirm that we are not calling the `reportDefinition/listAll` API when we are on the landing page? (not sure if it's the correct naming, lol)

davidcui1225:
> Do you confirm that we are not calling the `reportDefinition/listAll` API when we are on the landing page? (not sure if it's the correct naming, lol)

Yes, I think you are referring to the GET `../api/reporting/reportDefinitions` API which is only called once within a helper function `refreshReportDefinitionsTableContent` that is no longer used anywhere in `main`

zhongnansu (approved):

joshuali925 (approved):

PR #299, opened 2021-01-08
Keep Reporting menu in Nav Menu when switching Index Patterns

davidcui1225:
*Issue #, if available:*
N/A
*Description of changes:*
* Fix in-context menu so it does not disappear when switching index patterns in the `Discover` page
* Instead of checking just that the navMenu length is greater than 1 on Report source pages, checks to make sure the length matches based on the report source type
   * Dashboard nav menu length should be 4
   * Visualization nav menu length should be 3
   * Saved search nav menu length should be 5

### DIFF
--- a/kibana-reports/public/components/context_menu/context_menu.js
+++ b/kibana-reports/public/components/context_menu/context_menu.js
@@ -234,12 +234,34 @@ $(function () {
   locationHashChanged();
 });
 
+const isDiscoverNavMenu = (navMenu) => {
+  return navMenu[0].children.length === 5 && 
+    ($('[data-test-subj="breadcrumb first"]').prop('title') === 'Discover' ||
+    $('[data-test-subj="breadcrumb first last"]').prop('title') === 'Discover')
+}
+
+const isDashboardNavMenu = (navMenu) => {
+  return navMenu[0].children.length === 4 && 
+    ($('[data-test-subj="breadcrumb first"]').prop('title') === 'Dashboard')
+}
+
+const isVisualizationNavMenu = (navMenu) => {
+  return navMenu[0].children.length === 3
+    && ($('[data-test-subj="breadcrumb first"]').prop('title') === 'Visualize')
+}
+
+
+
 function locationHashChanged() {
   const observer = new MutationObserver(function (mutations) {
     const navMenu = document.querySelectorAll(
       'span.kbnTopNavMenu__wrapper > nav.euiHeaderLinks > div.euiHeaderLinks__list'
     );
-    if (navMenu && navMenu.length && navMenu[0].childElementCount > 1) {
+    if (navMenu && navMenu.length && (
+      isDiscoverNavMenu(navMenu) ||
+      isDashboardNavMenu(navMenu) ||
+      isVisualizationNavMenu(navMenu))
+    ) {
       try {
         if ($('#downloadReport').length) {
           return;
@@ -263,9 +285,9 @@ function locationHashChanged() {
   });
 }
 
-window.onhashchange = function () {
+$(window).one('hashchange', function( e ) {    
   locationHashChanged();
-};
+});
 /**
  * for navigating to tabs from Kibana sidebar, it uses history.pushState, which doesn't trigger onHashchange.
  * https://stackoverflow.com/questions/4570093/how-to-get-notified-about-changes-of-the-history-via-history-pushstate/4585031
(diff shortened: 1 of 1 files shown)


kibana-reports/public/components/context_menu/context_menu.js
```diff
@@ -234,12 +234,34 @@ $(function () {
   locationHashChanged();
 });
 
+const isDiscoverNavMenu = (navMenu) => {
+  return navMenu[0].children.length === 5 &&
```
zhongnansu, line 238: why is it always 5, 4, or 3. I feel it's kind of fragile to used `== (a fixed number)`

kibana-reports/public/components/context_menu/context_menu.js
```diff
@@ -234,12 +234,34 @@ $(function () {
   locationHashChanged();
 });
 
+const isDiscoverNavMenu = (navMenu) => {
+  return navMenu[0].children.length === 5 && 
+    ($('[data-test-subj="breadcrumb first"]').prop('title') === 'Discover' ||
+    $('[data-test-subj="breadcrumb first last"]').prop('title') === 'Discover')
```
zhongnansu, line 240: why there are 2 conditions being checked, while the other functions contain one.

davidcui1225: Because on the `Discover` page the Reporting menu will be present in the menu even if there is no Saved search selected, but the id of the `Discover` breadcrumb prop is different when there is a Saved search selected vs. when there isn't

kibana-reports/public/components/context_menu/context_menu.js
davidcui1225: It checks how many items are in the nav menu depending on what kind of report source it is- I can extract them to constants too

zhongnansu: can we remove this length check, since you have the title check？Or can we just use `length >= 0` or `length >= 1`? What's the downside you see

davidcui1225: That was the old check, which resulted in the issue. Basically it would not add the Reporting menu if the nav menu had length greater than 1, which is why the Reporting menu would not show up. But we need to not only check that the Nav menu is populated, but that it has the correct number of items

kibana-reports/public/components/context_menu/context_menu.js
zhongnansu: got it. So, this only happens with Discover, not the other two?

davidcui1225: Yeah- because on the base `Dashboard` and `Visualizations` screens it is just a list of the available ones, but you are not on the actual page with the nav menu

zhongnansu: Did you try the "create visualization/dashboard"? I think it leads to a similar scenario with `Discover`, with no saved object selected, but with a nav bar

kibana-reports/public/components/context_menu/context_menu.js
joshuali925: a better way might be to add an id to reports menu element, loop through navMenu and check for the id to determine whether reports menu is added. probably a todo item

kibana-reports/public/components/context_menu/context_menu.js
davidcui1225: Good catch- I will have to add those checks as well

kibana-reports/public/components/context_menu/context_menu.js
```diff
@@ -263,9 +285,9 @@ function locationHashChanged() {
   });
 }
 
-window.onhashchange = function () {
+$(window).one('hashchange', function( e ) {
```
joshuali925, line 288: > Yes, it has had no problems with switching in between so far. One thing I just found is that the Reporting menu does not show up in Discover until you select a Saved search though

Does it show up if you don't select a saved search but refresh the page?

joshuali925 (approved):
some comments

kibana-reports/public/components/context_menu/context_menu.js
davidcui1225: +1 - I think that's a good idea but it doesn't sound like a quick fix

kibana-reports/public/components/context_menu/context_menu.js
davidcui1225: I have added a fix for that issue as well, now it does show up on base Discover page

zhongnansu (approved):
LGTM, be careful when backporting to `7.9.1`

PR #313, opened 2021-02-05
Support customized server config

### LINKED ISSUE #98: Access kibana.yml from plugin to read server.host, server.port and server.basePath

zhongnansu:
*Issue #, if available:*
#98
This is the most reported issue so far. Users are not able to generate any report if they have customized Kibana sever setting(host, sever base path, port), or enable TLS.

Server config is set in kibana.yml

*Description of changes:*
- remove the hardcoded `localhost:5601` when building the complete url
- Dynamically retrieve `serverInfo`(which contains `port`, `host`, `protocol`), and `server.basePath` and pass them through route handler for validating the input url, compose complete url.
- Add one more puppeteer launch option `ignoreHttpsError: true` to support TLS-enabled kibana server https://github.com/puppeteer/puppeteer/blob/main/docs/api.md#puppeteerlaunchoptions
- Add one more param "path: server.basePath" to the cookie object, to support customized server.basePath when security is enabled.

### DIFF
--- a/kibana-reports/server/index.ts
+++ b/kibana-reports/server/index.ts
@@ -13,8 +13,34 @@
  * permissions and limitations under the License.
  */
 
-import { PluginInitializerContext } from '../../../src/core/server';
 import { OpendistroKibanaReportsPlugin } from './plugin';
+import { schema, TypeOf } from '@kbn/config-schema';
+import {
+  PluginInitializerContext,
+  PluginConfigDescriptor,
+  HttpServerInfo,
+} from '../../../src/core/server';
+
+export const configSchema = schema.object({
+  access: schema.object({
+    port: schema.number({ defaultValue: 5601 }),
+    basePath: schema.string({ defaultValue: '' }),
+  }),
+});
+
+export type KibanaReportsPluginConfigType = TypeOf<typeof configSchema>;
+
+export const config: PluginConfigDescriptor<KibanaReportsPluginConfigType> = {
+  exposeToBrowser: {
+    access: true,
+  },
+  schema: configSchema,
+};
+
+export type AccessInfoType = {
+  basePath: string;
+  serverInfo: HttpServerInfo;
+};
 
 //  This exports static code and TypeScript types,
 //  as well as, Kibana Platform `plugin()` initializer.
--- a/kibana-reports/server/index.ts
+++ b/kibana-reports/server/index.ts
@@ -14,28 +14,10 @@
  */
 
 import { OpendistroKibanaReportsPlugin } from './plugin';
-import { schema, TypeOf } from '@kbn/config-schema';
 import {
   PluginInitializerContext,
-  PluginConfigDescriptor,
   HttpServerInfo,
 } from '../../../src/core/server';
-
-export const configSchema = schema.object({
-  access: schema.object({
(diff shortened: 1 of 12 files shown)


kibana-reports/server/index.ts
```diff
@@ -13,8 +13,34 @@
...
+  }),
+});
+
+export type KibanaReportsPluginConfigType = TypeOf<typeof configSchema>;
+
+export const config: PluginConfigDescriptor<KibanaReportsPluginConfigType> = {
+  exposeToBrowser: {
+    access: true,
+  },
+  schema: configSchema,
```
joshuali925: what does this file do besides defining `KibanaReportsPluginConfigType`? it seems it's adding port and basepath config to the kibana.yml file? shouldn't be necessary since we are not reading from yml

davidcui1225 (approved):

kibana-reports/server/utils/validationHelper.ts
```diff
@@ -45,8 +45,14 @@ export const regexRelativeUrl = /^\/(_plugin\/kibana\/app|app)\/(dashboards|visu
 
 export const validateReport = async (
   client: ILegacyScopedClusterClient,
-  report: ReportSchemaType
+  report: ReportSchemaType,
+  basePath: String
 ) => {
+  report.query_url = report.query_url.replace(basePath, '');
+  report.report_definition.report_params.core_params.base_url = report.report_definition.report_params.core_params.base_url.replace(
+    basePath,
```
joshuali925, line 53: is query_url/base_url the full url with hostname? e.g. `localhost:5601/alh/app/...` would it replace `alh` in `localhost`

joshuali925 (approved):

kibana-reports/server/utils/validationHelper.ts
zhongnansu: `query_url` doesn't contain hostname

kibana-reports/server/index.ts
zhongnansu: This is something I forget to remove when I tried the plugin config approach. Will remove

PR #331, opened 2021-02-22
Add dynamic wait to allow page content to render

### LINKED ISSUE #330: Report empty if using plugins with client side calculations
When using plugins that do client side calculations, such as Kibana Enhanced Table, reports are empty as the screenshot is taken before the page has fully rendered.

itbm:
*Issue #, if available:*
Fixes #330 

*Description of changes:*
Add wait before screenshot to allow content to render

By making a contribution to this project, I certify that:

(a) The contribution was created in whole or in part by me and I
have the right to submit it under the open source license
indicated in the file; or

(b) The contribution is based upon previous work that, to the best
of my knowledge, is covered under an appropriate open source
license and I have the right under that license to submit that
work with modifications, whether created in whole or in part
by me, under the same open source license (unless I am
permitted to submit under a different license), as indicated
in the file; or

(c) The contribution was provided directly to me by some other
person who certified (a), (b) or (c) and I have not modified
it.

(d) I understand and agree that this project and the contribution
are public and that a record of the contribution (including all
personal information I submit with it, including my sign-off) is
maintained indefinitely and may be redistributed consistent with
this project or the open source license(s) involved.

### DIFF
--- a/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
+++ b/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
@@ -140,6 +140,9 @@ export const createVisualReport = async (
         `report source can only be one of [Dashboard, Visualization]`
       );
   }
+  
+  // wait for page content to render
+  await page.waitFor(10000);
 
   const screenshot = await page.screenshot({ fullPage: true });
 
--- a/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
+++ b/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
@@ -59,6 +59,24 @@ export const createVisualReport = async (
     ? DOMPurify.sanitize(header)
     : DEFAULT_REPORT_HEADER;
   const reportFooter = footer ? DOMPurify.sanitize(footer) : '';
+  
+  // add waitForDynamicContent function
+  const waitForDynamicContent = async(page, timeout = 30000, interval = 1000, checks = 5) => {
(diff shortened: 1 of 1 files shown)


zhongnansu (commented):
We try to avoid hardcode as mush as possible. Could you think of using any puppeteer APIs that doesn't require a interval as input? Something like waitUntil(), waitForNavigation. May need to look up their API doc as well

itbm:
I had looked at those originally, but could not see any able to take in to account scripts still modifying html after page load (at least not without adding a waitForSelector).

itbm:
@zhongnansu I have tested the other options and they all result in the blank plugin

zhongnansu:
@itbm Hi, I found someting on the internet which could be a possible generic solution for waiting for client side JS scripts to complete. 
https://stackoverflow.com/a/61304202
Also, I think `page.waitFor` is going to deprecate, maybe try `page.waitForTimeout()`?

itbm:
> @itbm Hi, I found someting on the internet which could be a possible generic solution for waiting for client side JS scripts to complete.
> https://stackoverflow.com/a/61304202

I think this kind of idea is probably ok for most use cases and adds some flexibility to extend the wait vs a simple `page.waitFor(x)`.

> Also, I think `page.waitFor` is going to deprecate, maybe try `page.waitForTimeout()`?

`page.waitForTimeout()` is not available in the included puppeteer version so cannot be used until the package is upgraded.

zhongnansu (approved):

joshuali925 (approved):
thanks for the fix

Issue #351, opened 2021-04-05
Time Range not using Commonly Used Ranges from Kibana Settings


iget-master:
Currently, the `time_range` component uses a constant from `report_settings_constants.tsx` to feed the `commonlyUsedRanges` property of `EuiSuperDatePicker` inside it.

This is not the default behavior across Kibana. The default behavior is to load this from `uiSettings` service, allowing the user to override the available options.

I'm proposing this change that is not a breaking change but a "fix" that makes this plugin follow the default behavior across Kibana.

iget-master:
Although my PR #352 fixes this, this plugin does not support relative dates (see #353). :disappointed:

zhongnansu:
This issue has been moved to https://github.com/opensearch-project/dashboards-reports/issues/41

PR #352, opened 2021-04-05
Use `commonlyUsedRanges` from Kibana Settings instead of a constant value

iget-master:
*Issue #, if available:* #351

*Description of changes:*

* Add the `uiSettings` service from `Data` plugin as dependency
* Removes unused dependencies (notifications and navigation)
* Change `time_range` component to use `commonlyUsedRanges` from `uiSettings` (following default behavior for `EuiSuperDatePicker` across entire kibana), allowing user to override defaults 
* Removed the constant that was used for `commonlyUsedRanges`

### DIFF
--- a/kibana-reports/public/application.tsx
+++ b/kibana-reports/public/application.tsx
@@ -20,14 +20,13 @@ import { AppPluginStartDependencies } from './types';
 import { OpendistroKibanaReportsApp } from './components/app';
 
 export const renderApp = (
-  { notifications, http, chrome, uiSettings }: CoreStart,
+  { http, chrome, uiSettings }: CoreStart,
   {  }: AppPluginStartDependencies,
   { appBasePath, element }: AppMountParameters
 ) => {
   ReactDOM.render(
     <OpendistroKibanaReportsApp
       basename={appBasePath}
-      notifications={notifications}
       http={http}
       uiSettings={uiSettings}
       chrome={chrome}
--- a/kibana-reports/public/application.tsx
+++ b/kibana-reports/public/application.tsx
@@ -20,8 +20,8 @@ import { AppPluginStartDependencies } from './types';
 import { OpendistroKibanaReportsApp } from './components/app';
 
 export const renderApp = (
-  { notifications, http, chrome }: CoreStart,
-  { navigation }: AppPluginStartDependencies,
+  { notifications, http, chrome, uiSettings }: CoreStart,
+  {  }: AppPluginStartDependencies,
   { appBasePath, element }: AppMountParameters
 ) => {
   ReactDOM.render(
@@ -29,7 +29,7 @@ export const renderApp = (
       basename={appBasePath}
       notifications={notifications}
       http={http}
-      navigation={navigation}
+      uiSettings={uiSettings}
       chrome={chrome}
     />,
     element
--- a/kibana-reports/public/components/app.tsx
+++ b/kibana-reports/public/components/app.tsx
@@ -47,7 +47,6 @@ export interface CoreInterface {
 
 interface OpendistroKibanaReportsAppDeps {
   basename: string;
-  notifications: CoreStart['notifications'];
   http: CoreStart['http'];
   // navigation: NavigationPublicPluginStart;
   chrome: CoreStart['chrome'];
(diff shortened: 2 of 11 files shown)


kibana-reports/public/components/app.tsx
```diff
@@ -47,10 +47,10 @@ export interface CoreInterface {
 
 interface OpendistroKibanaReportsAppDeps {
   basename: string;
-  notifications: CoreStart['notifications'];
   http: CoreStart['http'];
-  navigation: NavigationPublicPluginStart;
+  // navigation: NavigationPublicPluginStart;
```
zhongnansu, line 51: If navigation and notifications are not used anywhere, please remove them from code, instead of comment it out.
@joshuali925 @davidcui-amzn could you confirm that it's safe to remove those 2?

kibana-reports/public/types.ts
```diff
@@ -13,13 +13,16 @@
  * permissions and limitations under the License.
  */
 
-import { NavigationPublicPluginStart } from '../../../src/plugins/navigation/public';
+// import { NavigationPublicPluginStart } from '../../../src/plugins/navigation/public';
```
zhongnansu, line 16: same here

kibana-reports/public/types.ts
```diff
@@ -13,13 +13,16 @@
...
-export interface OpendistroKibanaReportsPluginSetup {}
+export interface OpendistroKibanaReportsPluginSetup {
+}
 
 // eslint-disable-next-line @typescript-eslint/no-empty-interface
 export interface OpendistroKibanaReportsPluginStart {}
 
 export interface AppPluginStartDependencies {
-  navigation: NavigationPublicPluginStart;
+  // navigation: NavigationPublicPluginStart;
```
zhongnansu, line 26: here

kibana-reports/public/application.tsx
```diff
@@ -20,16 +20,15 @@ import { AppPluginStartDependencies } from './types';
 import { OpendistroKibanaReportsApp } from './components/app';
 
 export const renderApp = (
-  { notifications, http, chrome }: CoreStart,
-  { navigation }: AppPluginStartDependencies,
+  { http, chrome, uiSettings }: CoreStart,
+  {  }: AppPluginStartDependencies,
```
zhongnansu, line 24: Do we need to `data` here?

zhongnansu (changes requested):

kibana-reports/public/components/app.tsx
zhongnansu: @davidcui-amzn can you take a look, and if it's safe to remove this, let's merge in the PR first

davidcui1225: Can confirm, `navigation` and `notifications` are not used. Please remove them instead of commenting out, as Zhongnan suggests above.

kibana-reports/public/application.tsx
davidcui1225: No, the parameter for `AppPluginStartDependencies` was `navigation` previously, which is not used

kibana-reports/public/application.tsx
```diff
@@ -20,16 +20,15 @@ import { AppPluginStartDependencies } from './types';
 import { OpendistroKibanaReportsApp } from './components/app';
 
 export const renderApp = (
-  { notifications, http, chrome }: CoreStart,
-  { navigation }: AppPluginStartDependencies,
+  { http, chrome, uiSettings }: CoreStart,
```
davidcui1225, line 23: If the `uiSettings` parameter is used to refer specifically to `commonlyUsedRanges`, can we use a more descriptive name for it? `uiSettings` is very generic

iget-master: uiSettings is the name of the service, I see no reason for not using the default name.

iget-master: It's the exact name used across kibana source code.

iget-master:
I'm moving this PR to https://github.com/opensearch-project/dashboards-reports with the requested changes. I think this repository will be cloded, right?

kibana-reports/public/application.tsx
davidcui1225: I see, no issue then

zhongnansu:
Moved to https://github.com/opensearch-project/dashboards-reports/pull/49

Issue #350, opened 2021-03-31
Low PDF Quality


iget-master:
I'm testing the PDF reporting functionality (my main reason to migrate to Opendistro) when I noticed that the generated PDF is just a screenshot of the page instead of a vector/text PDF. Also there's a issue with some fonts with wrong kerning.

I've reading the code and doing some tests, and looks viable to remove the screenshot part of the code:

https://github.com/opendistro-for-elasticsearch/kibana-reports/blob/98750bee4fcf64b1c4505d9f39e8b5908a95de3a/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts#L170-L177

So the quality is restored, but the font kerning issue persists. Doing some research, found [this issue on puppeteer](https://github.com/puppeteer/puppeteer/issues/2410) that suggests adding `--font-render-hinting=none` to args solves the problem. I did tested here, and the result is a perfect PDF.

Note that since the removed piece of code is responsible for adding the header and footer to the report, we need to rewrite the `composeReportHtml` to modify the original html instead of creating a new one.

I'm working on a pull request to solve this issue, but would be great to hear from someone that wrote the screenshot part the reasons for it.

zhongnansu:
@iget-master 
Hi that's an issue that has been brought up a lot. Honestly I forget why we made a deisgn choice at that time lol, but some time ago I did have a branch in my own repo that does some research on high-quality, text-selectable pdf. But I think at that time we just don't have enough time and resource to complete that and add to the release. Please take a look if you can find anything useful. The changes are mainly in one file,  and I mostly get it working. But I didn't think of the font issue at that time. And there are minor issues of header/footer rendering, which I documented in the TODOs.

Please let me know if you have a different approach and I am willing to discuss more.


https://github.com/zhongnansu/kibana-reports/blob/research-text-selectable-pdf/kibana-reports/server/routes/utils/visualReportHelper.ts#L113-L169

iget-master:
@zhongnansu your solution is similar to mine, except for the fix for wrong fonts. But now I noticed that I made it only for Dashboard as source, I'll test with visualization

zhongnansu:
This issue has been moved to https://github.com/opensearch-project/dashboards-reports/issues/45

PR #354, opened 2021-04-06
Improve quality of rendered PDFs

iget-master:
Also a issue with chrome font rendering was fixed

*Issue #, if available:* #350 

*Description of changes:*

1. Removed screenshot from visual report. Instead of it, we add the header and footer to the original Dashboard HTML
2. Added `font-render-hinting=none'` to chromium args to fix font rendering issues
3. Removed Query Bar from report since it's not relevant to end user. I considered adding an option on Report Definition to allow choice to remove it or not, but noticed that it will require changes to `report-scheduler` plugin, and I'm not used to Kotlin language :facepalm: 

Except for the point 3, this PR doesn't introduces any breaking changes.


kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
```diff
@@ -204,26 +204,44 @@ export const createVisualReport = async (
...
 
-  return $.root().html() || '';
-};
+const addReportHeader = async (page, header) => {
+  const headerHtml = fs.readFileSync(`${__dirname}/header_template.html`)
+    .toString()
+    .replace('<!--CONTENT-->', header);
+
+  await page.evaluate((headerHtml: string) => {
+    const dashboardViewport = document.querySelector('#dashboardViewport');
```
zhongnansu, line 223: Need to handle visulization as well. Also we have defined variables for the id selectors for dashboard/visualization, please use them instead.
```
SELECTOR.dashboard
SELECTOR.visualization
```

kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
```diff
@@ -131,19 +131,24 @@ export const createVisualReport = async (
...
-    }
+      /* istanbul ignore next */
+      () => {
+        // remove buttons
+        document
+            .querySelectorAll("[class^='euiButton']")
+            .forEach((e) => e.remove());
+        // remove query bar
+        document
+            .querySelectorAll("[class='globalQueryBar']")
```
zhongnansu, line 142: I think this `globalQueryBar` also includes the time range. We still need to display that in the pdf/png.

zhongnansu (changes requested):

zhongnansu:
@iget-master Thanks for contributing. Could you include a png of the final report in your PR description? I think that will make it easier for reviewers and people to refer.

zhongnansu:
Hi do you get a chance to continue working on this PR? I assume the major missing piece is to add support for `Visualization` as well?

It would be better that if you can post this PR to our new Opensearch repo. But let me know if you'd like this to merge in for OpenDistro. We can figure it out

zhongnansu:
No response, migrate to this PR #379

PR #356, opened 2021-04-06
Version bump to 1.13.2.0

zhongnansu:
*Issue #, if available:*

*Description of changes:*


kibana-reports/server/routes/utils/converters/backendToUi.ts
```diff
@@ -222,12 +222,15 @@ const getUiQueryUrl = (
   const timeFrom = moment(beginTimeMs).toISOString();
   const timeTo = moment(endTimeMs).toISOString();
   let queryUrl = `${baseUrl}?_g=(time:(from:'${timeFrom}',to:'${timeTo}'))`;
-  if (tenant) {
-    if (tenant === '__user__') {
+  if (tenant !== undefined) {
```
davidcui1225, line 225: What's this change? I don't think we planned to add this to the 1.13.2 release?

davidcui1225 (approved):

kibana-reports/server/routes/utils/converters/backendToUi.ts
joshuali925: this is from https://github.com/opendistro-for-elasticsearch/kibana-reports/pull/324 which zhongnan backported to 1.13 branch. this seems like a github issue to me, i don't think it should show up in diff..

joshuali925 (approved):

kibana-reports/server/routes/utils/converters/backendToUi.ts
zhongnansu: cherry-pick creates new commit id, and github diff I guess is not comparing the actual diff, instead if somehows uses commid id to compare, just a guess

joshuali925: git should be able to figure out commits that have same content with different hash, I think github uses 3 dot diff `git diff ours...theirs` to preview merge, but this result is weird. I merged locally and only version bumps are in the diff so shouldn't affect us

Issue #327, opened 2021-02-19
Error in CSV report


Artain:
## Version
Opendistro Elasticsearch 1.13.0 standalone linux
Opendistro Kibana 1.13.0 standalone linux

## Error description
If the saved discover table includes a null value the csv export will fail with the error message:
```
[08:16:09.781] [error][opendistroReportsKibana][plugins] Failed to generate report: TypeError: Cannot read property 'toString' of null
  log   [08:16:09.782] [error][opendistroReportsKibana][plugins] TypeError: Cannot read property 'toString' of null
    at sanitize (/srv/opendistroforelasticsearch-kibana/plugins/opendistroReportsKibana/server/routes/utils/dataReportHelpers.ts:227:18)
    at getEsData (/srv/opendistroforelasticsearch-kibana/plugins/opendistroReportsKibana/server/routes/utils/dataReportHelpers.ts:172:34)
    at convertEsDataToCsv (/srv/opendistroforelasticsearch-kibana/plugins/opendistroReportsKibana/server/routes/utils/savedSearchReportHelper.ts:266:18)
    at generateReportData (/srv/opendistroforelasticsearch-kibana/plugins/opendistroReportsKibana/server/routes/utils/savedSearchReportHelper.ts:151:10)
    at process._tickCallback (internal/process/next_tick.js:68:7)
 error  [08:16:07.104]  Error: Internal Server Error
    at HapiResponseAdapter.toError (/srv/opendistroforelasticsearch-kibana/src/core/server/http/router/response_adapter.js:132:19)
    at HapiResponseAdapter.toHapiResponse (/srv/opendistroforelasticsearch-kibana/src/core/server/http/router/response_adapter.js:86:19)
    at HapiResponseAdapter.handle (/srv/opendistroforelasticsearch-kibana/src/core/server/http/router/response_adapter.js:81:17)
    at Router.handle (/srv/opendistroforelasticsearch-kibana/src/core/server/http/router/router.js:164:34)
    at process._tickCallback (internal/process/next_tick.js:68:7)

```

hollden:
I have almost the same error on 1.12:
```
{"type":"log","@timestamp":"2021-02-18T07:56:03Z","tags":["error","plugins","opendistroReportsKibana"],"pid":21258,"message":"Failed to generate report by id: TypeError: Cannot read property 'settings' of undefined"}
[...]

PR #361, opened 2021-04-18
Improve sanity tests

uzhinskiy:
Pass processing if doc[field] == null
Based on https://github.com/opendistro-for-elasticsearch/kibana-reports/issues/327#issuecomment-796479152

*Issue #327 :*

*Description of changes:*
Calling doc [field] .toString () for null-values was throwing an error. Added check for null-value.


zhongnansu:
LGTM

zhongnansu (approved):
Thanks for the fix

davidcui1225 (approved):
Thanks for the fix- could you also add a small test case for this change as well?

zhongnansu:
No response, let's merge it in and add the test case by ourself

Issue #363, opened 2021-04-20
Missing fields in the generated file


uzhinskiy:
I am trying to create a CSV report based on a saved search that contains fields like **@timestamp, A, B, C.D, C.E, F**, etc. But the generated file only contains **@timestamp, A, B, F** - in the report all complex fields are missing. 
In fact - the fields that mapped as _object_ in my index are missed.
Why did this happen and how can I fix it?

Kibana v 7.10.0 from docker-image "amazon/opendistro-for-elasticsearch-kibana:1.12.0"

zhongnansu:
Could you be more specific about "complex fields", also it's better if you can provide some example, screenshot, dataset you used, anything

uzhinskiy:
For example:
```
   {
        "_index" : "my_index",
        "_type" : "_doc",
        "_id" : "789be064-0758-4245-8ca1-6570df3bb5a7",
        "_score" : 1.0,
        "_source" : {
          "status" : "alert",     -   simple field = string
          "issue_uid" : 123,   -   simple field = number
          "created_at" : "2020-09-21T05:02:48.000+03:00",    -   simple field = date
          "alerts_count" : 2,   -   simple field = number
          "labels" : {       -   complex field = object 
            "kubernetes" : "cluster.k8s",    -  part of complex field  -  will be missed from CSV-report
            "statefulset" : "sts1"  -  part of complex field  -  will be missed from CSV-report
          }
        }
    }

```

zhongnansu:
This issue has been moved to https://github.com/opensearch-project/dashboards-reports/issues/43

PR #367, opened 2021-05-14
Fix csv parsing function

joshuali925:
Signed-off-by: Joshua Li <joshuali925@gmail.com>

*Issue #, if available:*
- https://github.com/opensearch-project/dashboards-reports/issues/43
- #363 

*Description of changes:*


zhongnansu:
Can we add test case for this?

zhongnansu:
Why CI is not triggered for this PR

joshuali925:
> Can we add test case for this?

Added. Did we have memory issues with csv reports? I had to flatten the hits to take care of selecting columns like `a.b.c` from `{ 'a.b': { 'c': 'value' } }`, this might increase memory consumption

joshuali925:
https://github.com/joshuali925/kibana-reports/actions check CI here, i think github is having problem with PR triggers

zhongnansu:
> > Can we add test case for this?
> 
> Added. Did we have memory issues with csv reports? I had to flatten the hits to take care of selecting columns like `a.b.c` from `{ 'a.b': { 'c': 'value' } }`, this might increase memory consumption

maybe, like this issue described. https://github.com/opensearch-project/dashboards-reports/issues/34

joshuali925:
> maybe, like this issue described. [opensearch-project/dashboards-reports#34](https://github.com/opensearch-project/dashboards-reports/issues/34)

Got it. I don't think the issue can be resolved by just optimize memory usage while generating csv. We'll have to introduce some kind of external storage for processing data. So this PR shouldn't affect that issue

zhongnansu (approved):

zhongnansu:
> > maybe, like this issue described. [opensearch-project/dashboards-reports#34](https://github.com/opensearch-project/dashboards-reports/issues/34)
> 
> Got it. I don't think the issue can be resolved by just optimize memory usage while generating csv. We'll have to introduce some kind of external storage for processing data. So this PR shouldn't affect that issue

I think a hard limit on csv file size may be another approach, such as 10 MB

zhongnansu:
can we merge this in? @joshuali925

joshuali925:
missing an approval, @davidcui-amzn could you take a look please?

davidcui1225 (approved):

PR #379, opened 2021-06-09
increase PDF report quality and make text selectable

zhongnansu:
*Issue #, if available:*
#350 

*Description of changes:*
See original PR #354  for more description. I just addressed the comments myself and add to the original PR, since that PR is stale.
**CI/CD**: https://github.com/zhongnansu/kibana-reports/runs/2785964481
pdf with high quality example: 
[On_demand_report_2021-06-09T02_40_40.218Z_142b07a0-c8cc-11eb-afaf-7dde810f2d81.pdf](https://github.com/opendistro-for-elasticsearch/kibana-reports/files/6626640/On_demand_report_2021-06-09T02_40_40.218Z_142b07a0-c8cc-11eb-afaf-7dde810f2d81.pdf)


davidcui1225 (approved):

kibana-reports/server/routes/utils/visual_report/style.css
```diff
@@ -0,0 +1,211 @@
...
+
+.mde-preview .mde-preview-content pre {
+  padding: 16px;
+  overflow: auto;
+  font-size: 85%;
+  line-height: 1.45;
+  background-color: #f7f7f7;
+  border-radius: 3px;
+}
+.mde-preview .mde-preview-content pre code {
```
joshuali925, line 157: just curious are these styles taken from react-mde library? if so we might be able to include them instead of defining them again

joshuali925 (approved):
LGTM

kibana-reports/server/routes/utils/visual_report/style.css
zhongnansu: I believe so. But how can we import it from the library?

joshuali925: it should be `@import 'react-mde/lib/styles/css/react-mde-preview.css';` or `@import '~react-mde/lib/styles/css/react-mde-preview.css';`, but i tried both didn't work. Not sure if kibana or pdf generation did something special, maybe it's easier to import the css in typescript in `addReportStyle` and merge it with `visual_report/style.css`, i can look into it later

joshuali925: it works by just changing the [css variable](https://github.com/zhongnansu/kibana-reports/blob/ebdfe99556966d11a28e03e9f497130980a293bd/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts#L195-L196) to 
```js
const css = fs.readFileSync(`${__dirname}/style.css`).toString() + fs.readFileSync(`${__dirname}/../../../../node_modules/react-mde/lib/styles/css/react-mde-preview.css`).toString();
```

but seems the current way might be better.. since we don't rely on css path in the library

zhongnansu: Thanks for all the investigation. I do agree that rely on css path in lib is kind of tricky. So let's keep it the existing way till we have a better solution

PR #381, opened 2021-06-16
remove exit full screen button

badbybirth:
Hello,

I'm using your plugin to export images from my Kibana dashboard.
I'm trying to get as little clutter on the export as possible.
Using the plugin it is possible to export visualizations in full screen.
In that case however a small button to exit the full screen view is rendered.
This small change removes this button.

### DIFF
--- a/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
+++ b/kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
@@ -123,6 +123,10 @@ export const createVisualReport = async (
       document
         .querySelectorAll("[class^='euiHeader']")
         .forEach((e) => e.remove());
+      // remove exit full screen button
+      document
+        .querySelectorAll("[class^='dshExitFullScreenButton']")
+        .forEach(e => e.remove())
       // remove visualization editor
       if (reportSource === REPORT_TYPE.visualization) {
         document


zhongnansu:
Hi, thx for contributing to the project. We have migrated OpenDistro to Opensearch. The plan is and we are deprecating Opendistro, any new feature will be release in Opensearch. We'll review this PR, but could you also send a PR in Opensearch repo? 
https://github.com/opensearch-project/dashboards-reports

joshuali925 (approved):
looks good to me, thanks

badbybirth:
'll check it out, and probably do it tomorrow.

kibana-reports/server/routes/utils/visual_report/visualReportHelper.ts
```diff
@@ -123,6 +123,10 @@ export const createVisualReport = async (
       document
         .querySelectorAll("[class^='euiHeader']")
         .forEach((e) => e.remove());
+      // remove exit full screen button
+      document
+        .querySelectorAll("[class^='dshExitFullScreenButton']")
```
zhongnansu, line 128: I am interested in the use case here. The `existFullScreenButton` is not there if you don't use full-screen mode, and we can generate report from the normal dashboard page. When enable full-screen, the button shows but reporting will not be available from UI. I wonder how did you use full-screen mode to generate report, did you have some implementation by our own?

badbybirth: I'm trying to export a single visualization ad-hoc without any clutter.
The ad-hoc visualization export also still includes elements I don't want on there. (search bar etc) 

The plugin just loads any Kibana url you send it, renders it using an embedded chrome, takes a screenshot, wraps that image in a header and footer, renders it again in chrome, takes a screenshot again, and spits out that image.
I just manipulated the path that is sent to the endpoint which generates the actual image to be a full screen representation of the visualization I want, by enabling full-screen, hiding the panel's name, etc...
When enabling full screen like this however, the button is present.

I'm planning on using your plugin to generate an image ad-hoc of a visualization, from another system, so I can include them in the reports it generates.
I'm aware using your plugin for this is not ideal, but it beats basically writing a plugin that does almost the same.

I haven't looked into how you can extend the Kibana UI, perhaps it might be possible to add a report button of your there later, so users can actually export full-screen images, and you can use the same approach used here to hide the report button.

zhongnansu: I understand but I don't this suits our use case. You can still make the modification in your forked version and use it, but I don't think we should merge it into this repo. Because in our use case, Chrome doesn't have a chance to take a url which leads to the full-screen mode.

If you feel like the visulization report has many clutters. Please submit a issue and list the elements you don't want there.

@anirudha @davidcui-amzn @joshuali925 any idea?

zhongnansu:
No reply. Close for now

